PR #106, opened 2025-01-11
Fix hold_court.6000

ProZeratul:
option A & B affecting every elligible counties instead of only the one selected

> In fact this is just reverting the event as it was design originally (prior to 1.13 patch) I got mislead thinking this event was broken since the event text have not been updated ...

Time wasted :(


ProZeratul:
Game devs changed it in 1.15.0

PR #146, opened 2025-05-06
Fix opinions when offering or demanding courtiers or concubines

pharaox:
* Make opinions types when offering or demanding courtiers and concubines consistent (e.g. `annoyed_opinion`, `pleased_opinion`, etc.)
* Make sure that opinion gains from repeatedly offering courtiers or concubines don't stack by using a custom opinion type, similarly to gifts. I regard this as a bug fix, since without it the player can gain unlimited opinion with particular rulers, especially since you can often get the courtier back via "invite to court".
* Fix a few ck3-tiger errors in the newly added files


pharaox:
I realized I am missing tooltips warning about no opinion gain from repeated offers, so don't merge this yet. Will add them shortly.

PR #147, opened 2025-05-07
Dev/1161 merge

ProZeratul:
@pharaox This is the merge for the new patch that was released this morning.


pharaox (approved):

pharaox (changes requested):
Looking at their changes in `gfx/court_scene/scene_cultures/00_default_cultures.txt`, I have some concerns. I think what they did wouldn't work and we should remove 2 of their lines. See how I propose to do it in https://github.com/ProZeratul/CrusaderKings3UnofficialPatch/pull/148.

ProZeratul:
@pharaox Nice catch on this one

PR #149, opened 2025-05-07
Fix pool characters marriage malus for nomadic rulers

pharaox:
Fixes the following bug:

* Try to marry one of your courtiers to a pool character of your religion and culture living nearby.
* They get -100 malus because "you are a filthy horse lord". What?

I believe this malus is intended for rulers only, hence the government check. The check however also fails for pool characters, so an explicit `is_ruler = yes` check is needed.


ProZeratul:
Good catch

PR #157, opened 2025-05-17
Fix ck3-tiger errors

pharaox:
Fix and ignore some remaining ck3-tiger errors. Also change one of my previous fixes since I noticed it causes errors in `error.log`.


ProZeratul:
I don't see anything wrong thanks for this

PR #163, opened 2025-05-25
Fix mpo_hunt.0020

pharaox:
Fix issues with `mpo_hunt.0020` event ("A Mother's Love"):

* You get the Wolf Pelt artifact even if you fail to kill the wolf, or if you try to sneak past her
* Stress impact makes no sense, e.g. you gain a large amount of stress if you are Sadistic and decide to kill the wolf


events/activities/hunt_activity/mpo_hunt_events.txt
```diff
@@ -842,8 +854,12 @@ mpo_hunt.0020 = {
 				}
 			}
 		}
+		#Unop No pelt if you tried to sneak past the mother
+		hidden_effect = {
+			destroy_artifact = scope:newly_created_artifact
+		}
 		stress_impact = {
-			calm = medium_stress_impact_gain
+			content = medium_stress_impact_gain #Unop Was calm
```
ProZeratul, line 862: Can you also change the calm in ai_chance to content ?

events/activities/hunt_activity/mpo_hunt_events.txt
```diff
@@ -704,11 +704,12 @@ mpo_hunt.0020 = {
 		show_as_tooltip = {
 			scope:newly_created_artifact = { set_owner = root }
 		}
-		stress_impact = {
```
ProZeratul: Can you add the `hunt_animal_wolf_trigger = yes` to trigger of the event `mpo_hunt.0020`  so we are sure a wolf is a valid animal

ProZeratul (changes requested):

events/activities/hunt_activity/mpo_hunt_events.txt
pharaox: Done, also changed the other `ai_chance` to be consistent with `stress_impact`.

ProZeratul:
I don't really understand why you changed the wolf trigger, as it means, return true everywhere in the world **except** in the region world_africa_west **if** the county is **neither** a hill/forest/jungle/mountain

So actually your modification is more restrictive than the existing one

```
NAND = {
	geographical_region = world_africa_west
	OR = {
		terrain = hills
		terrain = forest
		terrain = jungle
		terrain = mountains
	}
}
```

ProZeratul:
While I understand your point but as the wolf seems to be the fallback beast, I'm not too sure this should be changed as it might have unforeseen consequences for now (and also that would mean we need to be sure this does not prevent some region in Africa to hunt.

Something that could be done is to make animal subspecies like it is done for deer.

pharaox:
> So I think this should be reverted for now

Done, removed `00_hunt_triggers.txt` entirely as this was the only change.

PR #165, opened 2025-05-26
Fix falconry animal type selection if there is a sighting

pharaox:
If there is a "sighting" in the province, the hunt event animal type selection logic should pick the sighting animal as the default `animal_type` (suggested as first and obvious choice), and only resort to random game selection if there is no sighting. This is ok for regular hunts, see `hunt.500`.

For falconry, the logic in the event `hunt.0530` misses an else, and so random game is selected even if there is a sighting. As a result, the player doesn't get the increased success chance from the sighting, unless the randomly selected animal happens to be the sighting one.


ProZeratul:
Indeed !

PR #171, opened 2025-05-31
Fix legitimacy loss & gain from marriage if nomadic

pharaox:
Fixes the following issues with legitimacy loss & gain from marriage:

* There is logic that was (presumably) intended to apply legitimacy penalties if the recipient marries a lowborn `secondary_actor`. It was written in a way that it never applied to anything.
* With MPO, there is new logic that was (again presumably) intended to add legitimacy if you marry a highborn courtier of another nomadic ruler with higher dominance. Instead, it is written in such a way that you get 100 legitimacy even if you are not marrying yourself, and even if the courtier in question is lowborn, so you can really get high amounts of legitimacy for nothing.

I had to make some assumptions about the original intentions, since the code is so buggy they are not entirely obvious, but I hope I got them correctly.

### DIFF
--- a/common/scripted_effects/00_marriage_interaction_effects.txt
+++ b/common/scripted_effects/00_marriage_interaction_effects.txt
@@ -75,6 +75,10 @@
 				}
 			}
 		}
+	}
+	#Unop Rework the section below for the recipient, as it was (presumably) intended
+	# For the actor, this = scope:secondary_recipient is always false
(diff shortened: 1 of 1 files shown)


ProZeratul:
I think you nailed it, at first I was wondering if this was just some kind of duplicate code from a merge conflict on their side.

PR #173, opened 2025-05-31
Fix mess_tent building that was missing as a potential upgrade after …

ProZeratul:
…a camp party


ProZeratul:
https://forum.paradoxplaza.com/forum/threads/mess-tent-does-not-upgrade-as-a-result-of-camp-revelry.1761344/

PR #175, opened 2025-05-31
Dev/summon fair courtiers fix

ProZeratul:


ProZeratul:
https://forum.paradoxplaza.com/forum/threads/summon-fair-courtiers-only-summon-1-even-though-the-tooltip-say-3.1760683/

PR #182, opened 2025-06-13
Fix "Employ Executioner" option for nomadic rulers

pharaox:
Fix "Employ Executioner" option of "Negotiate Release" interaction being shown for nomadic rulers, even though they can't employ an executioner at all and this court position is not even shown.


ProZeratul:
Good catch

PR #183, opened 2025-06-16
Fix "Summon Wealthy Visitors" decisions

pharaox:
The "Summon Wealthy Visitors" decisions have a number of issues:

* The merchant is sent traveling to the employer's location, rather than their domicile location. Since there is no availability check in the decision, the employer could be traveling or leading an army at this time.
* Upon arriving, the events check whether the employer is available, and the merchant has arrived in their location. If the latter fails, the merchant is teleported to their location, and the event is re-triggered in 2 to 4 months. However, by the time the event fires the employer can be again in an entirely different place, resulting in the merchant "chasing" the employer sometimes for quite a while.
* The journey takes quite long, so the merchant pool character may not arrive for a number of reasons: arriving as a guest in another court, dying "in an accident" (actually being pruned by the game), being selected by a pool selector as a minor ruler, etc.
* If the merchant dies after the first journey is over, the player won't get notified about their death.
* Even though there is a list of origin cities, there is also a check for `development_level >= 30` which only true for Constantinople at game start, so many players only see Constantinople being selected.
* The Artifact Materials modifier gained after the second journey **disables** the *Commission Artifact* decision rather than enabling it. I guess nobody complained because nobody got that far.
* There are 2 decisions, "Local Offers" and "Paiza" depending on whether the Paiza system is established, for whatever reason with different cooldowns, 8 and 4 years.

I changed the above so that:

* The merchant is sent to the employer's domicile location.
* Upon arriving, if the employer is available, the merchant is immediately teleported to their domicile location (which is the employer's location as well in this case), and the event takes place. If not, the event is re-triggered in 2 to 4 months as before. The merchant will still have to wait until the employer is available, but will not have to chase them so the event will happen sooner.
* The merchant is now prevented from being selected by pool selectors, participating in events, arriving as a guest, being pruned, or becoming inspired, until the second journey is over. Also, if they die for whatever reason, the player will be notified appropriately.
* The development level check has been changed to `development_level >= 10` so most origin cities are now possible at any time.
* Added an additional modifier to origin city selection that make cities closer to root's domicile location more likely to be selected.
* Fixed the Artifact Materials modifier to enable the *Commission Artifact* decision and all its options.
* The cooldown for both decisions is now 4 years.

In addition, I fixed a few other issues that `ck3-tiger` reported in the newly added files.

### DIFF
--- a/common/scripted_effects/09_dlc_mpo_scripted_effects.txt
+++ b/common/scripted_effects/09_dlc_mpo_scripted_effects.txt
@@ -1505,7 +1505,7 @@ mpo_find_suitable_merchant_effect = {
 		1 = { #Merv
 			trigger = {
 				title:c_bukhara = {
-					development_level >= 30
+					development_level >= 10 #Unop Increase the chance for other origins
 					NOT = {
 						holder.top_liege = {
 							government_has_flag = government_is_nomadic
@@ -1516,6 +1516,10 @@ mpo_find_suitable_merchant_effect = {
 			modifier = {
 				add = title:c_bukhara.development_level
 			}
+			#Unop Favor origins closer to root's domicile location
+			modifier = {
+				add = title:c_bukhara.title_province.unop_paiza_merchant_distance_modifier_value
+			}
 			title:b_bukhara = {
 				save_scope_as = merchant_origin
 			}
@@ -1523,7 +1527,7 @@ mpo_find_suitable_merchant_effect = {
 		1 = { #Nishapur
 			trigger = {
 				title:c_nishapur = {
-					development_level >= 30
+					development_level >= 10 #Unop Increase the chance for other origins
 					NOT = {
 						holder.top_liege = {
 							government_has_flag = government_is_nomadic
@@ -1534,6 +1538,10 @@ mpo_find_suitable_merchant_effect = {
 			modifier = {
 				add = title:c_nishapur.development_level
 			}
+			#Unop Favor origins closer to root's domicile location
+			modifier = {
+				add = title:c_nishapur.title_province.unop_paiza_merchant_distance_modifier_value
+			}
 			title:b_nishapur = {
 				save_scope_as = merchant_origin
 			}
@@ -1541,7 +1549,7 @@ mpo_find_suitable_merchant_effect = {
 		1 = { #Khotan
 			trigger = {
 				title:c_khotan = {
-					development_level >= 30
+					development_level >= 10 #Unop Increase the chance for other origins
 					NOT = {
 						holder.top_liege = {
 							government_has_flag = government_is_nomadic
@@ -1552,6 +1560,10 @@ mpo_find_suitable_merchant_effect = {
 			modifier = {
 				add = title:c_khotan.development_level
 			}
+			#Unop Favor origins closer to root's domicile location
+			modifier = {
+				add = title:c_khotan.title_province.unop_paiza_merchant_distance_modifier_value
+			}
 			title:b_khotan = {
 				save_scope_as = merchant_origin
 			}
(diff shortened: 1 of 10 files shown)


common/scripted_effects/09_dlc_mpo_scripted_effects.txt
```diff
@@ -1505,7 +1505,7 @@ mpo_find_suitable_merchant_effect = {
 		1 = { #Merv
 			trigger = {
 				title:c_bukhara = {
-					development_level >= 30
+					development_level >= 10 #Unop Increase the chance for other origins
```
ProZeratul, line 1508: Is there a reason for 10 ? (Just wondering why you just didn't made it 15 (so nerf by 50%) ?)

common/scripted_effects/09_dlc_mpo_scripted_effects.txt
```diff
@@ -1676,16 +1712,20 @@ mpo_find_suitable_merchant_effect = {
 				}
 			}
 			modifier = {
-				add = title:c_baghdad.development_level
+				add = title:c_venezia.development_level #Unop: Venice
```
ProZeratul, line 1715: Didn't you heard of baghdad of Venezia ? 😆  (good catch 👍 )

common/scripted_triggers/00_courtier_guest_management_triggers.txt
```diff
@@ -0,0 +1,836 @@
...
+	#Isn't too good for host
+	NAND = {
+		OR = {
+			any_claim = { tier = tier_empire }
+			any_close_family_member = { highest_held_title_tier = tier_empire }
+		}
+		$HOST$ = { highest_held_title_tier <= tier_county }
+	}
+
+	#Unop Prevent paiza merchants from arrving as guests
```
ProZeratul, line 328: Minor typo

ProZeratul (changes requested):

localization/replace/english/unop_fixed_keys_l_english.yml
```diff
@@ -257,4 +257,7 @@
  # relation > target_relation
  agent_events.3041.b.tt.success: "[target_relation.GetFirstNameNoTooltip] will stay"
  # relation > target_relation
- agent_events.3041.b.tt.failure: "[target_relation.GetFirstNameNoTooltip] is free to leave"
\ No newline at end of file
+ agent_events.3041.b.tt.failure: "[target_relation.GetFirstNameNoTooltip] is free to leave"
+
+ mpo_call_for_merchants_decision.merchant_will_arrive_generic: "A wealthy foreigner will travel to your court, leading a caravan bound for the well-developed, settled cities. Upon their return, you may receive the following benefits:\n$BULLET_WITH_TAB$[artifact|E] creation materials that unlock the #V $commission_artifact_decision$#! [decision|E]\n$BULLET_WITH_TAB$Armaments for your [horde|E] and [archer_cavalry|E]"
```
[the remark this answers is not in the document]
pharaox: Because it's duplicated, you see both 20% and 40% in one case and nothing in the other.

common/scripted_triggers/00_courtier_guest_management_triggers.txt
pharaox: Where?

common/scripted_effects/09_dlc_mpo_scripted_effects.txt
ProZeratul: I'll merge it as is, we'll see if some people raise concern

ProZeratul:
@pharaox I was going to merge it, but it has conflict with the now merged pre-release files. From what I can see, the conflict resolution shouldn't be difficult.

pharaox:
@ProZeratul Resolved the conflicts now, ready to merge.

PR #188, opened 2025-06-29
Restore fixes in birth_events.txt

pharaox:
Restore fixes in `birth_events.txt`. These fixes existed in an older version of Unop, before the GitHub repo was created, but were somehow lost. I stumbled upon them because I have copied them to another mod I maintain (Carnalitas). They seem sensible to me so I propose to restore them. I did some minor edits but preserved the original behavior.


ProZeratul:
Indeed there are still some fixes that got left behind with RtP patch :( 

Thanks for this

PR #190, opened 2025-06-30
Dev/fix witch nickname

ProZeratul:


common/scripted_effects/00_nickname_effects.txt
```diff
@@ -526,7 +526,7 @@ assign_random_nickname_effect = {
 			100 = {
 				trigger = {
 					has_trait = witch
-					NOT = { faith = { has_doctrine = doctrine_witchcraft_accepted } }
+					NOT = { faith = { has_doctrine_parameter = witchcraft_accepted } } #Unop: Check doctrine param instead since witchcraft can be virtuous
```
pharaox, line 529: There is another similar usage of `has_doctrine = doctrine_witchcraft_accepted` in the same file on line 4541, we should change it as well I believe.

I also see similar checking of doctrine instead of parameter in 2 events files, in all cases I think the same fix should be applied:

* `events/dlc/bp1/bp1_filippa_yearly_events.txt`
* `events/travel_events/travel_events_filippa.txt`

pharaox (changes requested):

PR #191, opened 2025-06-30
Dev/fix tournament script values

ProZeratul:


common/script_values/04_ep2_tournament_values.txt
```diff
@@ -891,6 +892,10 @@ contest_type_aptitude_horse = {
 			else_if = {
 				limit = { has_trait = education_martial_prowess_4 }
 				add = 1
+			}			
+			else_if = { #Unop: Added Level 5
+				limit = { has_trait = education_martial_prowess_5 }
+				add = 1
```
pharaox, line 898: Besides the missing level 5, also the values added seem wrong - they are always 1 irrespective of level. Other similar values scale the aptitude with the level. This particular one may be different, since I see a comment that the final values should be between 0 and 5.

ProZeratul: Yes I left 1 since the comment on this value is 

```
# 	The end result of this number MUST be between 0 and 5, otherwise the game
# 	will crash due to being used in the UI to allocate a segmented progress bar
```

Since it's capped at 5 at the end, this still could be modified safely. Do you think level 5 should go to 3 out of 5 for this ? (since all others are 2 except level 1 that only get the default value)

pharaox (approved):

ProZeratul:
I'm reverting this, I got things wrong, in fact the only thing that exists for the prowess level 5 are the YML entries. So there was no issue there.

PR #193, opened 2025-06-30
Dev/playdate

ProZeratul:


common/activities/activity_types/playdate.txt
```diff
@@ -17,12 +19,21 @@
...
 		age >= 4
+		age < less_than_two_years_to_adulthood_value #Consistency!
+		highest_held_title_tier > tier_barony
+		trigger_if = {
+			limit = { highest_held_title_tier = tier_county }
+			custom_tooltip = {
+				text = unop_playdate_invalid_independent_count
+				is_independent_ruler = no
+			}
+		}
```
pharaox, line 36: I see you preserved the vanilla logic and only added a tooltip, but I don't get the idea - so if you are a count and independent, you can't invite peers, while if you are not independent, you can. Why? Shouldn't it be the opposite (you can only do it if you are independent), or not checked at all? I thought it could be because there would be noone to invite, but looking at the guests handling code, this also doesn't seem to be the case.

ProZeratul: Yes the idea to display the info for the user came from several report on reddit/pdx forum where they asked why they couldn't setup this activity, I didn't change the requirements at all.

I agree that an independant count might get no one, I'll recheck this

pharaox (approved):
Approving this since I am not sure my concerns are valid.

PR #194, opened 2025-07-02
Fix ep3_laamp_decision_event.1031 & ep3_laamp_decision_event.1035 opt…

ProZeratul:
…ions CD begin way more than intended

https://forum.paradoxplaza.com/forum/threads/i-know-this-one-option-has-wrong-cooldown.1822033/


pharaox (approved):
LOL

PR #198, opened 2025-07-05
Pharaox fix ck3 tiger errors 5

pharaox:
Fix ck3-tiger errors in recently added files.

For hairstyles, I believe we should replace the missing templates with the `_afro` ones, and only for sub-saharan add all except `_afro`. In this way, we would be adding modifiers for all templates that actually exist.

`ck3-tigger` now supports suppressing errors via comments. I used this in one place where I verified that the error is a false positive. If you agree, we could use this style for suppressing such errors in the future as well.


ProZeratul:
I have no issue with using the new pragma of tiger, just we'll have to make sure it doesn't make future merge harder when updating to new release of the game

PR #205, opened 2025-07-09
Internal fix: is_valid_to_hire_court_position_type is run on the empl…

ProZeratul:
…oyee scope not employer, and is not what we want to check

### DIFF
--- a/common/scripted_triggers/00_court_position_triggers.txt
+++ b/common/scripted_triggers/00_court_position_triggers.txt
@@ -161,8 +161,10 @@ court_position_does_not_already_have_a_job_trigger = {
 # $CHAR$ is the character you wish to appoint
 can_appoint_char_to_court_position = {
 	save_temporary_scope_as = hiring_liege
-	#can_employ_court_position_type = $COURT_POS$ #Unop: Replaced this with the line below to work even if the char is not in our court
-	is_valid_to_hire_court_position_type = $COURT_POS$
+	OR = {  #Unop: We need to have an OR since can_employ_court_position_type will return false if we have the maximum amount of character for a given slot (like 2 bodyguard)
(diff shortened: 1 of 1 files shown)


common/scripted_triggers/00_court_position_triggers.txt
```diff
@@ -161,8 +161,10 @@ court_position_does_not_already_have_a_job_trigger = {
 # $CHAR$ is the character you wish to appoint
 can_appoint_char_to_court_position = {
 	save_temporary_scope_as = hiring_liege
-	#can_employ_court_position_type = $COURT_POS$ #Unop: Replaced this with the line below to work even if the char is not in our court
-	is_valid_to_hire_court_position_type = $COURT_POS$
+	OR = {  #Unop: We need to have an OR since can_employ_court_position_type will return false if we have the maximum amount of character for a given slot (like 2 bodyguard)
+		can_employ_court_position_type = $COURT_POS$
+		employs_court_position = $COURT_POS$
+	}
```
pharaox, line 167: The change itself looks correct. I struggled a bit to understand the difference between `is_valid_to_hire_court_position_type` and `would_be_valid_for_court_position` (we use this below). It seems the second includes the first but also checks the employer, and therefore should be used everywhere where the first is currently still used. What do you think?

pharaox (approved):
lgtm

common/scripted_triggers/00_court_position_triggers.txt
ProZeratul: **is_valid_to_hire_court_position_type** also check that the character is in a court, would_be_valid_for_court_position does not, so you'll probably want to use **would_be_valid_for_court_position** everywhere

PR #208, opened 2025-07-10
Review bp3_journey_events.txt checks for CP

ProZeratul:
This file have a lot of errors in the way of checking CP available positions

Main issues:

**is_valid_to_hire_court_position_type** is an **employee** trigger
**employs_court_position** and **any_court_position_holder** are **employer** triggers
So the events trigger are mixing employee/employer triggers for root

--

- [ ] Add comments everywhere where unop_employer_can_have_cp is called
- [ ] In game Tests

### DIFF
--- a/events/dlc/mpo/mpo_migration_travel_events.txt
+++ b/events/dlc/mpo/mpo_migration_travel_events.txt
@@ -914,6 +914,16 @@ mpo_migration_travel_events.0050 = {
     }
 
     immediate = {
+        every_courtier = {
+            limit = { is_eligible_courtier_0050 = yes }
+            add_to_list = unop_potential_courtier
+        }
+        location = {
+            every_character_in_location = {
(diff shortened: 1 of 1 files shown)


pharaox (changes requested):
When searching for `is_valid_to_hire_court_position_type`, I see a number of places where this trigger is used outside of `bp3_journey_events.txt` in vanilla. All of them seem to be incorrect, following the same pattern of checking the employer rather than the employee.

* common/scripted_effects/00_journey_effects.txt
* events/travel_events/travel_events_bp3.txt
* events/dlc/mpo/mpo_nomads_flavour_events_oltner.txt

And I see virtually no places where this trigger is used correctly. I am wondering, could it be that the usages are actually correct, and the docu is wrong? Have you tested this, does the trigger work indeed as documented?

If the docu is correct, then let's replace all occurrences of this trigger with our improved trigger also in the above files.

PR #210, opened 2025-07-10
Dev/fix scheme friendship modifier

ProZeratul:
https://forum.paradoxplaza.com/forum/threads/ritualized-friendship-adds-20-or-40-days-to-scheme-length.1836232/


pharaox (approved):
lgtm

PR #212, opened 2025-07-11
Dev/fix legends

ProZeratul:
https://forum.paradoxplaza.com/forum/threads/ai-cant-start-legends.1759835/


pharaox (approved):
Another amazing fix :-) I saw the forum post, but have you tested it, does it really work?

ProZeratul:
@pharaox Yes it does work, here is from a game that only got the papacy with a legend, let some time passe at speed 5 and you get everyone to start creating ones.

image

PR #223, opened 2025-07-18
Scale employer_has_treasury_to_offer_job_trigger with the tier to match contract rewards

### LINKED ISSUE #222: In the beginning as a LAAMP, you would be offered far too many criminal contracts
In some of my LAAMP playthroughs, I noticed that at the beginning I was offered almost only criminal contracts, despite my camp purpose not being Freebooters. E.g. as Scholars, I was getting tons of "Flog Talismans" contracts, since that's the only Learning criminal contract, and almost nothing else, except the occasional "Settle Boundary Dispute".

I investigated this, and I think the reason is that non-criminal contract check `employer_has_treasury_to_offer_job_trigger` (all except "Settle Boundary Dispute"), which does essentially `scope:employer.gold >= 50`, while criminal contracts don't. As a result, if you happen to be an area where barons and counts are relatively poor, and since you don't have the level of fame required to get contracts from richer dukes and kings, you only get criminal and "Settle Boundary Dispute" contracts.

The contract reward scales with the tier, but the above check doesn't. So one possible fix would be to scale the check with the tier as well, ensuring that barons and counts may have less than 50 gold.

pharaox:
Fixes #222

### DIFF
--- a/common/scripted_triggers/00_laamp_triggers.txt
+++ b/common/scripted_triggers/00_laamp_triggers.txt
@@ -449,7 +449,25 @@ valid_laamp_basic_accept_only_trigger = {
 employer_has_treasury_to_offer_job_trigger = {
 	trigger_if = {
 		limit = { is_ai = no }
-		scope:employer.gold >= 50
+		#Unop Scale treasury requirement to be half of the standard contract reward for the tier
+		# See task_contract_tier_gold_multiplier_value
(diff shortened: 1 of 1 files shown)


ProZeratul:
I'm merging this, as I said in the related issue, the gold taken from AI for contract should not impact them too much (or at all)

PR #226, opened 2025-07-18
Fix "Offer Courtier" and "Offer Concubine" for LAAMPs

pharaox:
The interactions "Offer Courtier" (newly introduced in 1.16) and "Offer Concubine" (existing before 1.16) can be sent by landless adventurers to rulers in their diplo range. I believe this is an omission, these interactions should be restricted only to nearby rulers, similarly to most other interactions when the actor is a landless adventurers.

### DIFF
--- a/common/character_interactions/00_marriage_interactions.txt
+++ b/common/character_interactions/00_marriage_interactions.txt
@@ -3218,6 +3218,19 @@ offer_concubine = {
 				allowed_more_concubines = yes
 			}
 		}
+		#Unop Limit range for landless adventurers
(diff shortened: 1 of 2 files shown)


ProZeratul:
You meant outside of their diplo range right ?

pharaox:
> You meant outside of their diplo range right ?
I meant that LAs shouldn't use their diplo range for these interactions, instead they should use the much smaller "LAAMP range" that is being used in most other interactions.

ProZeratul:
Agreed with this, there might be more interaction that lack this check

PR #227, opened 2025-07-18
Fix "Demand Conversion" hook not increasing acceptance for LAAMPs

pharaox:
I noticed that if you use the "Demand Conversion" interaction on courtiers as a landless adventurer, you can spend a hook if you have one, but this doesn't increase the acceptance at all. I found out that this is explicitly disallowed in `religion_demand_conversion_default_modifier`.

From my perspective, this doesn't make sense. LAs should be able to spend a hook and get the benefits if they are demanding conversion from their own followers, or demanding conversion from house members as the house head. The only case where this check makes sense is in the context of the "Attempt to Convert" interaction (attempting to convert any ruler if you have a certain perk), because they are already getting a (smaller) benefit in the interaction itself.

I modified the above modifier accordingly.

### DIFF
--- a/common/scripted_modifiers/00_religion_scripted_modifiers.txt
+++ b/common/scripted_modifiers/00_religion_scripted_modifiers.txt
@@ -1,0 +1,537 @@
+﻿#Scripted Modifiers based on a character's Religion or Faith.
+religion_scaled_virtuous_traits_modifier = {
+	modifier = {
+		add = {
+			add = $BASE_VALUE$
+			multiply = $CHARACTER$.num_virtuous_traits
+		}
+		$CHARACTER$ = {
+			num_virtuous_traits > 0
+		}
+	}
+}
+
+religion_scaled_sinful_traits_modifier = {
+	modifier = {
+		add = {
+			add = $BASE_VALUE$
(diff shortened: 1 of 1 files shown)


pharaox:
I have 1.13.2 (latest RtP patch) saved, the code looks exactly the same there, and before RtP there are no landless adventurers, or "Attempt to Convert". So it's likely that the check is there from the very beginning of RtP and is therefore "by design". I am still struggling to understand why would it be designed like this, and a sloppy fix to prevent the hook to be used twice in the "Attempt to Convert" interaction seems like the most likely explanation. Do you have other ideas?

We could also a consider another approach - changing the `religion_demand_conversion_default_modifier` modifier to always apply a hook bonus of 20, instead of 50, for LAs, consistent with the bonus applied in "Attempt to Convert", and removing the bonus from the interaction.

I guess if we are not sure we should just leave it as it is. But then, we should consider changing the affected interactions to disable the hook entirely for LAs, since it's currently rather confusing. What do you think?

ProZeratul:
No I think this solution to be the most "in-line" with what exists in game currently. I was saying this just in case since I couldn't figure why they would have done it this way.

ProZeratul (approved):

PR #231, opened 2025-07-19
Fix tributary or suzerain not being able to join wars where their suzerain / tributary is a primary defender

### LINKED ISSUE #230: As a triburary or suzerain, I can't join wars where my suzerain / tributary is a primary defender
From Steam comments:

> Captain Smollett 13 Jul @ 7:01pm 
> 00_alliance.txt line 3621 and below - remove NOT for both tributary and suzerain, it is supposed to let you join defensive wars of your tributary/suzerain (judging by the comment and the tooltip in game) yet the NOT in both prevents that

Indeed, these 2 `NOT`s, starting at line **3261**, are misplaced and should be removed. To fix this properly, also the following changes are needed:

* Change the tooltip in the second case to talk about suzerain, not tributary.
* Change the corresponding checks in `can_be_picked` to match the checks in `is_valid_showing_failures_only`.

The correct logic should always be:

* As a suzerain, you can join a war where a direct or an indirect tributary is a primary defender.
* As a tributary, you can join a war where your direct suzerain (but not their suzerain or above)  is a primary defender

pharaox:
Fixes #230

### DIFF
--- a/common/character_interactions/00_alliance.txt
+++ b/common/character_interactions/00_alliance.txt
@@ -3267,13 +3267,13 @@ join_war_interaction = {
 				}
 				custom_tooltip = {
 					text = join_war_interaction_recipient_tributary_no_defensive_war
-					NOT = {
+					#NOT = { #Unop Revert the check to work as intended
 						scope:recipient = {
 							any_character_war = {
 								primary_defender = scope:recipient
 							}
 						}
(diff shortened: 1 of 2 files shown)


ProZeratul:
Also shouldn't this be placed there ? 
image

PR #237, opened 2025-07-22
Dev/misc yml

ProZeratul:
For the virtue fix: https://forum.paradoxplaza.com/forum/threads/instill-virtue-works-on-0-year-old-rulers.1851066/


pharaox:
@ProZeratul Great yml fixes, could we perhaps separate the yml fixes and the Wet Nurse fix in 2 different PRs, since they are very different in nature?

PR #240, opened 2025-07-28
Dev/fix mystic healing

ProZeratul:
Seen there : https://forum.paradoxplaza.com/forum/threads/mystic-trait-level-1-is-treated-as-level-2-for-mystic-physician-treatment-success-chance.1852600/


pharaox (approved):
lgtm

Issue #65, opened 2024-12-21
Landless adventurers can donate gold and ask for blessing despite being of a Hostile or Evil faith


pharaox:
See https://forum.paradoxplaza.com/forum/threads/landless-adventurers-can-donate-gold-and-ask-for-blessing-despite-being-of-a-hostile-or-evil-religion.1722183/

![s1-donation](https://github.com/user-attachments/assets/3c887e93-dfe0-4805-b83c-e526f5a81f8c)

pharaox:
@ProZeratul The fix that I propose is to simply grey out the first 2 options (donation and blessing), unless you are either the same religion, the same faith, or perhaps of a faith that is "compatible" (not considered hostile or evil by the temple's faith, and you don't consider them hostile or evil), with an appropriate tooltip.
E.g. as an Orthodox Christian, you could donate or ask for blessing in a Catholic church, but not in a mosque.

PR #252, opened 2025-08-03
Fix "Path - Attainment" event penalizing you for speaking the local language

### LINKED ISSUE #251: Speaking the local language decreases the success chance in "The ... Path — Attainment", instead of increasing it
In the "The ... Path — Attainment" event (ep3_laamp_decision_event.1103) speaking the local language or converting to a faith of the same religion should increase the chance of success. Instead, these decrease the chance.

Image
Image

In the above screenshot, the player character speaks the local language and gets penalized for it. With the fix:

Image

pharaox:
Fixes #251 and #65

### DIFF
--- a/events/dlc/ep3/ep3_laamp_decision_events.txt
+++ b/events/dlc/ep3/ep3_laamp_decision_events.txt
@@ -14401,6 +14401,11 @@ ep3_laamp_decision_event.1043 = {
 				}
 			}
 		}
+		#Unop Limit this to same religion
+		trigger = {
+			faith.religion = scope:visiting_location.faith.religion
+		}
+		show_as_unavailable = { always = yes }
 		if = {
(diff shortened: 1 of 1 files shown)


events/dlc/ep3/ep3_laamp_decision_events.txt
```diff
@@ -14482,6 +14487,8 @@ ep3_laamp_decision_event.1043 = {
 	option = {
 		name = ep3_laamp_decision_event.1043.b
 		trigger = {
+			#Unop Limit this to same religion
+			faith.religion = scope:visiting_location.faith.religion
```
ProZeratul, line 14491: Shouldn't we restrict the same religion to only zealous character and allow it for same faith (at least if astray?)

PR #294, opened 2025-11-01
Fix a few unrelated issues

pharaox:
Fix a few unrelated issues:

* Revert the change to `events/dlc/ce1/epidemic_events.txt` in an attempt to prevent a CTD. That peasant is a throwaway character so it probably doesn't matter if they are properly infected.
* Use correct scope in `common/character_interactions/00_grant_titles_interaction.txt` (ck3-tiger complained about this one).
* Use `?=` correctly in `common/scripted_effects/10_dlc_tgp_natural_disaster_scripted_effects.txt` (the game complains about this line on startup)


ProZeratul:
Lgtm

PR #295, opened 2025-11-04
Fix MPO court positions and other issues

pharaox:
Fix multiple issues with MPO court positions (`common/court_positions/types/00_mpo_court_positions.txt`):
* Their salaries are a mess, it seems the file got corrupted by search & replace.
* The AI never hires them, because they lack `ai_position_score`.

Also:
* Wrong position references in `common/court_positions/types/00_mandala_court_positions.txt`.
* Missing entries for the new court positions in `court_position_does_not_already_have_a_job_trigger`.
* Wrong trigger scope in `events/dlc/ep3/ep3_laamp_decision_events.txt`.

### DIFF
--- a/common/scripted_triggers/00_court_position_triggers.txt
+++ b/common/scripted_triggers/00_court_position_triggers.txt
@@ -199,15 +199,6 @@ court_position_does_not_already_have_a_job_trigger = {
 			has_court_position = camp_priest_camp_officer
 			has_court_position = laureate_camp_officer
 
-			#Unop: Missing entries for TGP court positions
-			has_court_position = grand_preceptor_court_position
-			has_court_position = grand_mentor_court_position
-			has_court_position = grand_guardian_court_position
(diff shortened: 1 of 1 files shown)


common/scripted_triggers/00_court_position_triggers.txt
```diff
@@ -199,6 +199,15 @@ court_position_does_not_already_have_a_job_trigger = {
...
 
+			#Unop: Missing entries for TGP court positions
+			has_court_position = grand_preceptor_court_position
+			has_court_position = grand_mentor_court_position
+			has_court_position = grand_guardian_court_position
+			has_court_position = fire_dragon_engineer_court_position
+			has_court_position = keeper_of_the_harem_court_position
+			has_court_position = court_guru_court_position
+			has_court_position = khlon_glan_court_position
+
```
ProZeratul: All of them are already here, but IDK why they didn't add them at the end instead (I was confused too at first)

PR #300, opened 2025-11-08
Fix Unop unlocking unlimited kingdom-level holy wars

### LINKED ISSUE #299: Unop unlocks unlimited kingdom-level holy wars for all rulers
From Steam comments

> Sledjer 17 hours ago 
> Is it intended that this mod unlocks unlimited kingdom-level holy wars? Seems a bit cheaty to have them without the downsides of the cultural tradition.

This looks like a bug in Unop. This check: https://github.com/ProZeratul/CrusaderKings3UnofficialPatch/blob/b9a65c45377f05386e80f02265d1791c6517e481/common/casus_belli_types/00_religious_war.txt#L1246 is inside the OR, causing a major religious war to be allowed if the flag is not set, rather than disallowing it if the flag is set. It should be moved after the OR.

pharaox:
Fixes #299

I reproduced the above issue and tested the fix in both cases - without and with the `bonus_major_holy_war` flag.

### DIFF
--- a/common/casus_belli_types/00_religious_war.txt
+++ b/common/casus_belli_types/00_religious_war.txt
@@ -1240,10 +1240,18 @@ major_religious_war = {
 			}
 			custom_description = {
 				text = declared_major_religious_war_flag_tt
-				OR = {
-					has_character_flag = bonus_major_holy_war
-					NOT = { has_character_flag = declared_major_religious_war_flag }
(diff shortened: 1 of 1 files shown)


ProZeratul:
Thanks for this.
Edit: Just to add context, this was done to allow you to perform another religious war in the case where you war got invalidated for whatever reason

PR #306, opened 2025-11-12
China wall project fix

still222:
https://forum.paradoxplaza.com/forum/threads/improve-great-wall-project-bugged.1865712
Save I used for testing included in the bug report.

My initial idea was wrong, so I reopened this request with a different change.

To show the "Improve the Great Wall" special project, game looks for all baronies with not-fully upgraded walls in hegemon's realm. Problem is that some of this baronies do have a wall, but don't have any holdings (empty). It confuses the special project later and, as a result, it doesn't show that barony as available which results in unfinishable special project.

Each barony that's available for upgrade is directly mentioned in the special project code and instead of fixing each and every one, I would rather make a check for showing the special project in the first place.

### DIFF
--- a/common/great_projects/types/00_great_project_types.txt
+++ b/common/great_projects/types/00_great_project_types.txt
@@ -210,14 +210,19 @@ mandala_capital_01 = {
 	invite_interaction = request_great_project_contribution_interaction
 
 	allowed_contributor_filter = {
+		owner = yes
 		tributaries = yes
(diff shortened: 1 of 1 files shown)


still222:
Bug in the link still exist in 1.18.1

It could be easier to compare with the 1.18.1 file directly, if you want to know changes. 
There is 1 line I added in comparison to vanilla

PR #309, opened 2025-11-12
Region change for the "Far From Home" achievement

still222:
A bit of a personal change. Achievement requires to have a capital on the "Islands in the Indian Ocean" Region. But actual region is very limited and includes only pre-1.18 islands. 

I used this wikipedia map: https://upload.wikimedia.org/wikipedia/commons/d/d3/Indian_Ocean-CIA_WFB_Map.png , to include a lot of the new islands to the region. Because region instance could be overwritten by a separate file, I decided to just cut the change out of the original.


ProZeratul:
Hello sorry for the late response, I don't really think we should tweak achievements, this is a really sensible topic in the community and I don't want people rioting on the fact that we are making things easier.

I'm closing as I think we should only touch achievements if there are issues with them in the most strict sense. (But if I think you have a really valid point)

PR #320, opened 2025-11-30
Fix ck3-tiger errors, add exclusions to ck3-tiger.conf

pharaox:
Fix some errors reported by ck3-tiger, and add an exclusion list for most errors reported due to tiger not yet being updated to CK3 1.18 (to be removed later when it is).

I got it to 0 errors and 79 warnings, mostly `strict-scopes`. Getting down to 0 would require more work, but it's a start.


pharaox:
Added 2 more commits that fix mainly `scopes` warnings. Still 98 warnings remaining, mainly `strict-scopes`.

pharaox:
Added 2 more commits that fix mainly strict-scopes warnings. Still 79 warnings remaining, mainly strict-scopes.

PR #331, opened 2025-12-23
Fix more ck3-tiger warnings

pharaox:
* Fix some error.log warnings, for `root` being used values when it's an activity, and for `has_house_power_parameter` used with an incorrect scope (you also have this in your PR, I commented on it).
* Fix some issues in `01_event_backgrounds.txt` (the others are ignored).
* Fix remaining ck3-tiger warnings and update ck3-tiger.conf.

ck3-tiger.conf:
* Update `ck3-tiger.conf` to disable some more false positives.
* Reorder it into 3 sections: "False Positives", "Ignored", and "False positives due to tiger not yet updated to 1.18", and ensure every exclusion is in its proper place.
* Remove exclusions that are no longer needed
* Fix a few genuine issues that were previously ignored.

The number of ck3-tiger errors / warnings should be now 0.

Once tiger is updated to 1.18, we should remove all exclusions below `# Ignore false positives because not yet updated to 1.18` and try again.

I noticed that you removed some `ck3-tiger` from `#Unop` comments, so the comments in this PR don't have it. Let me know what you prefer.

### DIFF
--- a/ck3-tiger.conf
+++ b/ck3-tiger.conf
@@ -46,7 +46,10 @@ filter = {
 		# Don't print reports that are likely false positives
 		confidence >= Reasonable
 
+		########################################################################
 		# False positives
+		########################################################################
+
 		NAND = { # max in random is likely valid
 			key = validation
 			text = "`max = ` is only for `ordered_` lists, `random_list`, and `duel`"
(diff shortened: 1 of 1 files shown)


common/event_backgrounds/01_event_backgrounds.txt
```diff
@@ -10218,7 +10218,7 @@ tour_arrival_turned_away = {
 		}
 		reference = "gfx/interface/illustrations/activity_splash_screens/tour_arrival_asia_suspicious.dds"
 		video_mask = "gfx/interface/video/tour_arrival.bk2"
-		environment = "environment_ep2_tour_arrival_asia_suspicious"
+		environment = "environment_ep2_tour_arrival_indian_suspicious" #Unop ck3-tiger
```
ProZeratul: Looks like this one should be environment_tgp_courtyard_asia

PR #334, opened 2025-12-28
Prevents AUH faiths from creating holy orders through coronation 

still222:
I like holy orders. But the game is clearly designed to prevent far-eastern faith from creating such. They don't have names and use ugly generic ones instead.

Usually it is impossible to create one, but when playing I surprisingly found one in my land (after noticing a decision to boot them out). I was curious and decided to check all the ways you can create a holy order. There is very few ways to do so: 1 is from generic decision, which is obviously blocked, 2 are specific for the Jomsvikings and Hashishins.

The only other way comes from a coronation event that creates a new holy order, and of course Paradox completely forgot to add AUH support to it, which allows you (or ai) to spawn holy orders for eastern faiths. I've added check, mimicking one from the decision to prevent player/ai from doing it.

### DIFF
--- a/events/activities/coronation_activity/coronation_events_1.txt
+++ b/events/activities/coronation_activity/coronation_events_1.txt
@@ -8694,7 +8694,10 @@ coronation_events.1032 = {
 
         trigger = {
         	exists = scope:barony
-        	NOT = {
+        	NOR = {
+				faith = {	#Unop: Prevent immaterial harmony faiths from access to holy orders.
+					has_doctrine = special_doctrine_immaterial_harmony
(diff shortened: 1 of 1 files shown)


still222:
By the way, small effect to fix my save (not needed in the main patch):
```
	effect = {
		every_religion_global = {
			every_faith = {
				limit = {
					has_doctrine = special_doctrine_immaterial_harmony
					any_faith_holy_order = { always = yes }
				}
				every_faith_holy_order = {
					leader = { destroy_title = primary_title }
				}
			}
		}
	}
```

Issue #332, opened 2025-12-27
Scheme error log spam on loading active save


still222:
I'm more interested if this is a widespread issue in the first place.

I've noticed that just loading my ~200 years save, even before unpause spams thousands of errors, mainly from "slander_scheme" and a few others schemes.

One of the main (but not the only) reasons for error spam was `valid_agent = { is_valid_agent_standard_trigger = yes }` line, which looks like a hardcoded piece of tech.
As far as I understood it tries to evaluate agents for some old bugged schemes, and it constantly increasing amount of these errors with time.

I fixed it for myself, but would rather verify if it affect others before attaching the fix to patch. I will attach my save but it's made with a couple of graphical mods. It should load mostly correctly with just the Unop.

still222:
https://drive.google.com/file/d/1FwrzAxMsXnuz6b3S4oyAs8a7i02zLpFe/view?usp=sharing
Save file for testing, 2.5k lines of error log just on load.

pharaox:
@still222 How did you fix this, if the trigger is a builtin? You could open a PR with your changes and mark it as Draft so we could take a look. 
This trigger used to be a script trigger before 1.18, it seems in 1.18 they decided to make it a builtin. The code of the old trigger can still be found in older versions. Unfortunately we can't know what changes Paradox did to it in 1.18 so it would be risky to replace it.
Have you tried opening an issue with Paradox in the forums?

pharaox:
Closing this, see comments to PR. If you think we should do something about it please reopen.

PR #341, opened 2025-12-29
Scheme error spam on load

still222:
YOLO attempt to fix error spam on load:
https://github.com/ProZeratul/CrusaderKings3UnofficialPatch/issues/332

Now "fix" is extremely ambitious tone as I didn't fixed every load error. For me, the most prevalent error type was:
```
[E][jomini_script_system.cpp:303]: Script system error!
  Error: is_valid_agent_standard_trigger trigger [ Owner not defined ]
  Script location: file: common/schemes/scheme_types/slander_scheme.txt line: 50 (slander:valid_agent)
```
My suggested patch fixes this error. I was playing with the patch since publishing an issue topic, and it seems stable enough.

But as it is a fix directly to the slander scheme, I have no idea why it was so buggy in the first place, unlike other schemes.
Also, while it fixes my most prevalent error, there is also multiple other schemes (including slander again) causing a slightly milder error spam from the `agent_join_chance` block, which I couldn't fix:
```
[E][jomini_script_system.cpp:303]: Script system error!
  Error: scope:owner trigger [ Failed context switch ]
  Script location: file: common/scripted_modifiers/00_scheme_scripted_modifiers.txt line: 1313 (ai_agent_join_chance_hostile_general_suite_modifier[hash#3042561583])
    file: common/schemes/scheme_types/slander_scheme.txt line: 45 (slander:agent_join_chance)
``` 
I couldn't prevent it from evaluating in the first place, like I did with the `is_valid_agent_standard_trigger`, and modifiers themselves were a little to large to fix them without understanding what's even happening with these specific schemes.

My list of naughty schemes that causes error spam from the `agent_join_chance`:
1. slander_scheme
2. promote_scheme
3. expand_power_base_scheme

### DIFF
--- a/common/schemes/scheme_types/slander_scheme.txt
+++ b/common/schemes/scheme_types/slander_scheme.txt
@@ -47,7 +47,15 @@
 		ai_agent_join_chance_basic_suite_modifier = yes
 		ai_agent_join_chance_hostile_general_suite_modifier = yes
 	}
-	valid_agent = { is_valid_agent_standard_trigger = yes }
+	valid_agent = {
+		trigger_if = {
+			limit = { exists = scope:owner }
+			is_valid_agent_standard_trigger = yes
+		}
+		trigger_else = {
+			always = no
+		}
+	}
 	agent_groups_owner_perspective = { courtiers guests scripted_relations peer_vassals vassals family }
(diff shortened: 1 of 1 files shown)


pharaox:
Closing this, see previous comments. If you think we should do something about it please reopen.

PR #342, opened 2025-12-29
Less harsh pruning

still222:
I did see bug reports for that issue when 1.18 was released, but now I've encountered it myself:
https://forum.paradoxplaza.com/forum/threads/vanilla-population-control-may-target-wrong-people-i-found-that-more-than-13-rulers-that-disappeared-without-a-trace.1870219

The only wife of my heir disappeared without a trace, which more-or-less meant she was pruned by the overcrowd management system. Which was a bummer, as it was a great wife and I had to find another to replace her.

I've decided to make it a little less overzealous over pruning my close family and, well here's the fix.

Now, with how often that pruning is called, I was really nervous over including heavy triggers. I played with it for awhile, didn't really noticed any bad effects on the performance, but still.

### DIFF
--- a/common/character_interactions/06_ep3_interactions.txt
+++ b/common/character_interactions/06_ep3_interactions.txt
@@ -11602,8 +11602,7 @@ house_head_request_interaction = {
 			}
 		}
 		is_shown = {
-			scope:actor.domicile ?= { has_domicile_building_or_higher = cabinet_of_curiosities_01 }
-
+			scope:actor.domicile ?= { has_domicile_parameter = house_head_can_ask_for_artifacts } #Unop Use domicile parameter to support all domicile buildings
 		}
 		flag = artifact
 		localization = ASK_FOR_ARTIFACT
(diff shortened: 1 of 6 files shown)


pharaox:
As a next step, we could examine the above events to see what causes the issue. I already examined 2 of them.
* `bp2_adult_education.0009:after`: The teacher_1 and teacher_2 scopes are initialized when the activity starts, and are conservatively disposed of with `silent_disappearance_ai_if_created_effect`. However, since the activity takes some time, they may be selected as barons during it, which fails the check. We could add a `is_ruler = no` check to the event.
* `travel_danger_events.1006:after`: The guide scope is initialized in immediate and could select an existing pool character. If this character happens to be a ruler's relative or spouse (there are a lot of wandering former rulers), this fails the check. We could modify the event to only disappear created characters.

pharaox:
I also noticed that there are occurrences of `death_disappearance` outside of the death management effects, 2 in `common` and 22 in `events`. These are mostly not in `after`, but as part of the event plot. If a ruler, their consort, or their relative disappears like this, this would be valid in most cases.

This also means that at least some instances of the issue described in https://forum.paradoxplaza.com/forum/threads/vanilla-population-control-may-target-wrong-people-i-found-that-more-than-13-rulers-that-disappeared-without-a-trace.1870219/ could actually be "works as designed" - except of course that since the same death reason is used, the player is confused and frustrated. We could address this by using different death reasons for the 24 cases above, even introducing Unop death reasons if needed. 

Note that there is really no "population control system" in vanilla, in the same sense as the population control mod. The vanilla game simply tries to avoid leaving temporary created or pool characters lingering around, which depends on the correct usage of the above 5 effects in hundreds of places. Looking at the results collected, it mostly does it correctly, except for a relatively few events.

PR #351, opened 2026-01-02
Harem keeper fixes

### LINKED ISSUE #350: Keeper_of_the_harem court poistion bugs
While working on a mod to rework concubines, I've noticed that `keeper_of_the_harem_court_position` has quite a few bugs.

1. First of all pretty mild one:
`keeper_of_the_harem_select_palace_beauties.tt` loc states that it triggers once a year. But in fact `keeper_of_the_harem_court_position_task_select_palace_beauties` has *on_monthly* effect instead. I've checked 2 languages I know, and loc strings both mention yearly effect. I played with this court position for awhile and have a feeling on_monthly effect is really out of place and too 'powerful' as they jingle with concubines too fast (potentially 12 new concubines per year). I'm pretty sure it was intended to be `on_yearly` effect.

2. I saw reports on PDX forums that it's impossible to get a high aptitude for harem keeper, thought report were for 1.18.0. Personally I found quite a lot of "good" aptitude candidates and I think it is WAI, even if a little harsh.

3. `keeper_of_the_harem_court_position_task_select_palace_beauties`: line 252 very early in the `on_monthly` bloc: it has "any_consort" trigger and a little late "random_consort" effect. This is incorrect, `_consort` in this parts of code means concubines+spouses. Spouses definitely should NOT be affected by this task, it will cause all kind of issues if scope:liege has an old wife.

4. Same task, `random_consort` (which should be `random_concubine` as I stated earlier) should include in limits previous checks. Currently it finds if there is suitable for pruning consort (which should be suitable concubine), and if it was successful, tasks just prunes random old consort, without respecting previous limits (so it could prune lovers, etc).

still222:
Fixes #350

### DIFF
--- a/common/court_positions/tasks/00_harem_manager_tasks.txt
+++ b/common/court_positions/tasks/00_harem_manager_tasks.txt
@@ -244,12 +244,12 @@ keeper_of_the_harem_court_position_task_select_palace_beauties = {
 		}
 	}
 
-	on_monthly = {
+	on_yearly = {	#Unop: Making it the same to the original loc description and reducing the rates of new concubines.
 		scope:liege = {
 			if = {
 				limit = {
 					number_of_fertile_concubines < max_number_of_concubines
-					any_consort = {
+					any_concubine = {	#Unop: changing to concubine from consorts
 						is_ai = yes
 						is_designated_diarch = no
 						OR = {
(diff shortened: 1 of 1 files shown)


pharaox (approved):
Everything makes sense here, thanks!

PR #343, opened 2025-12-30
Analyze and fix silent disappearance issues

pharaox:
* Collect statistics about the issue described in #342 in order to understand and analyze it.
* Add fixes for all occurrences for 130 years after 1066.

PR #352, opened 2026-01-02
Fix ck3-tiger errors and warnings

pharaox:
* Fix new ck3-tiger errors and warnings  after updating tiger to 1.14
* Remove the `False positives due to tiger not yet updated to 1.18` section from `ck3-tiger.conf`, and add new false positives.

The number of ck3-tiger warnings and errors is now 0.

Some of the fixes are duplicated with fixes I saw in #327, but I added them anyway. If you decide to merge that one first, I will rebase and adapt.

Depends on #343 (this one should be merged first as it's used as a base).


ProZeratul:
I'll merge it as is, will fix the missing space directly, thanks for this.

PR #355, opened 2026-01-04
Fix Sinitic Syncretism, add rf_sinitic where appropriate

### LINKED ISSUE #349: Sinitic Syncretism doesn't work
From Steam comments

>  Müslibär [has Crusader Kings III] 28 Dec @ 3:12pm  
> Could you take a look at this one? I organized my religion and they still hate me.
[www.reddit.com/r/CrusaderKings/comments/1oqde3d/what_is_the_point_of_sinitic_syncretism/](http://www.reddit.com/r/CrusaderKings/comments/1oqde3d/what_is_the_point_of_sinitic_syncretism/)

Indeed, Sinitic Syncretism seems to be a mess. The main issue is that Sinitic faiths use  `eastern_hostility_doctrine` instead of `sinitic_hostility_doctrine`, and therefore `hostility_override_sinitic_hostility_doctrine` doesn't do anything. 

The fix proposed in Reddit changes it to `hostility_override_eastern_hostility_doctrine`, this works, however it would make it equivalent to Eastern Syncretism (will cover all Eastern faiths).

Another approach would be to replace `eastern_hostility_doctrine` with `sinitic_hostility_doctrine` for Sinitic faiths and make sure all places in code the former is used are changed as well. I found 8 places in 3 files. Also a larger adaptation of `common/religion/doctrines/30_core_tenets.txt` may be required to account for 4 hostility doctrines instead of 1.

I am more inclined to do the easy and safe Reddit fix as the other change would require more careful testing. @ProZeratul What do you think?

pharaox:
Fixes #349

As I suspected, this turned out to be a relatively big change, affecting 17 different files.

* Add `special_doctrine_is_sinitic_faith` for Sinitic faiths to tolerate Sinitic Syncretism faiths.
* Change all syncretism doctrines in `30_core_tenets.txt` so that Synitic Syncretism and its hostility doctrine are properly taken into account.
* Change `rf_sinitic`, `confucianism` and `taoism` to use `sinitic_hostility_doctrine` and  `special_doctrine_is_sinitic_faith`.
* Add `sinitic_hostility_doctrine` in most places where only `eastern_hostility_doctrine` was used previously.
* Add `tenet_sinitic_syncretism` in most placs where only `tenet_eastern_syncretism` was used previously.
* Add `rf_sinitic` in some places where only `rf_eastern` was used previously, and it was obvious that `rf_sinitic` should be added. I left a few places where I suspected it might be needed unchanged because I wasn't sure.

I considered whether some existing faiths with Eastern Syncretism should be changed to Sinitic Syncretism and reviewed them, but found none appropriate. Even those that were heavily China based (like Mingism) are more appropriately Eastern Syncretism.

I tested that:
* It's possible to add the Sinitic Syncretism doctrine while reforming and it has correct cost.
* Sinitic Syncretism is now Astray / Astray with Sinitic, but Hostile / Hostile with Eastern, and Eastern Syncretism is still Astray / Astray with Eastern, but Hostile / Hostile with Sinitic.
* Running the game for 10 years after 1066 did not produce error logs mentioning "sinitic" or "eastern".
* There are no `ck3-tiger` warnings for the newly added files or my changes.

@ProZeratul It would be great if you could test this independently on your side as it's quite a change.

Note that everything works as it should only if you start a new game. With existing games, nothing should be broken, but Sinitic faiths will consider Sinitic Syncretism faiths as Hostile instead of Astray, because their initial doctrines are written to the save.

### DIFF
--- a/localization/replace/english/unop_fixed_keys_l_english.yml
+++ b/localization/replace/english/unop_fixed_keys_l_english.yml
@@ -398,6 +398,8 @@ contest_events.1133.wound_enemy:0 "\n\nAs the sparks fly, I spot an opportunity,
  culture_parameter_waterworks_additional_bonuses: "The [GetBuilding('waterworks_01').GetTypeName|V] line of [buildings|E] have additional bonuses in [counties|E] of this [culture|E]"
  # Removed "and be able to produce children for longer" since it's not really the case
  game_concept_life_expectancy_desc:0 "$game_concept_life_expectancy$ is the number of years a character is expected to live before dying of natural causes.\n\nA high $game_concept_life_expectancy$ will make characters visibly age slower."
+ # Replaced rf_eastern with rf_sinitic below
+ sinitic_hostility_doctrine_desc: "The following religions are Sinitic:\n[GetReligionFamily('rf_sinitic').GetReligions]"
 
  # Use the [SelectLocalization( IsLandlessAdventurer(lover).. to not have duchy's appearing when the lover is an adventurer (disable need more test)
  #spymaster_task.0311.secret_lover.lustful:0 "\n\n[secret_holder.GetSheHe|U] has been engaged in carnality with [lover.GetTitledFirstName], running [secret_holder.GetHerHis] hands across the [SelectLocalization( IsLandlessAdventurer(lover), activity_adventure_owner, lover.GetTitleTierNamePossessive)] body, locking lips in all manner of ways... Oh, the scandal if this was to become public knowledge!"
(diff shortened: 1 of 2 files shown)


ProZeratul:
Hello, looks good so far, just some note:

Could you add this to the fixed keys:
` sinitic_hostility_doctrine_desc: "The following religions are Sinitic:\n[GetReligionFamily('rf_sinitic').GetReligions]"`

Change the **doctrine_background_icon = with core_tenet_banner_eastern.dds** with core_tenet_harmonious_society.dds

And lastly, during my test I saw some missing YML:

image

I've search and saw those keys potentially missing:

```
 sinitic_syncretic_recipient_opinion_active_opinion_name: "$faith_syncretic_opinion_active_control$"
 sinitic_syncretic_actor_opinion_active_opinion_name: "$faith_syncretic_opinion_active_control$"
 sinitic_syncretic_recipient_opinion_active_county_opinion_name: "$faith_syncretic_popular_opinion_active_control$"
 sinitic_syncretic_actor_opinion_active_county_opinion_name: "$faith_syncretic_popular_opinion_active_control$"
```

If you can add all this I think it'll be good for merge

ProZeratul (commented):
See my comments here: https://github.com/ProZeratul/CrusaderKings3UnofficialPatch/pull/355#issuecomment-3730896399

ProZeratul (changes requested):
See my comments here: https://github.com/ProZeratul/CrusaderKings3UnofficialPatch/pull/355#issuecomment-3730896399

ProZeratul:
Indeed I got I wrong on the banner, I was looking at the tenet icon my bad.

Merging as the YML changes are now done

PR #356, opened 2026-01-04
Fix house_bloc_desire_to_join_modifiers and invite_to_house_bloc_interacion

pharaox:
There are bugs in `house_bloc_desire_to_join_modifiers` that are affecting my current Japan game. While trying to fix them, ck3-tiger warned about a number of other issues that I am also fixing here. See the comments on the individual changes for more details.

### DIFF
--- a/common/character_interactions/10_tgp_japan_interactions.txt
+++ b/common/character_interactions/10_tgp_japan_interactions.txt
@@ -602,11 +602,11 @@ invite_to_house_bloc_interaction = {
 			}
 			add = 150
 		}
-		house_bloc_desire_to_join_modifiers = { # General modifiers
-			INVITER = scope:actor
-			JOINER = scope:recipient
-			MULTIPLIER = -1
-		}
+		#house_bloc_desire_to_join_modifiers = { # General modifiers #Unop Commented out as it doesn't make sense
+		#	INVITER = scope:actor
+		#	JOINER = scope:recipient
+		#	MULTIPLIER = -1
+		#}
  		modifier = {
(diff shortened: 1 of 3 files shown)


common/character_interactions/10_tgp_japan_interactions.txt
```diff
@@ -0,0 +1,2629 @@
...
+				exists = house.house_confederation
+				any_confederation = {
+					has_leading_house = yes
+					leading_house.house_head.top_liege = scope:actor.top_liege
+					leading_house = { has_same_house_power_as = scope:actor.house }
+				}
+			}
+			add = 150
+		}
+		#house_bloc_desire_to_join_modifiers = { # General modifiers #Unop Commented out as it doesn't make sense
```
ProZeratul: Why the comment on this one ? at a glance it's checking all sort of things 🤔

pharaox: In `ai_will_do`, you are checking the recipients desire to join actor's block **with a multiplier of -1** (so reversed) - so the more the recipient is reluctant to join, the more the actor is keen to invite them? Does this make any sense?
Not to mention that the multiplier applies to only some of the modifiers, so using a value different than 1 actually results in a complete mess.
If you prefer I can revert this one, but I strongly suspect it's wrong and likely cripples the AI.

ProZeratul: Do you think it's worth that I take some time to copy this one into a dedicated unop modifier that would handle all cases properly ?

PR #367, opened 2026-01-15
Fix tribal raiding capture intent not having modifiers applied

ProZeratul:
Fix https://forum.paradoxplaza.com/forum/threads/capture-raid-intent-does-not-actually-increase-capture-chance-for-tribal-rulers-code-error-found.1895244/


pharaox (approved):
lgtm

PR #368, opened 2026-01-15
Fix gold not being taken/given since the effect is now in a show_as_tooltip in purchase_land_interaction_effect

ProZeratul:
Since purchase_land_interaction_effect is no longer taking the gold, we need to actually take the gold from those who called it

### DIFF
--- a/common/scripted_effects/07_dlc_ep3_scripted_effects.txt
+++ b/common/scripted_effects/07_dlc_ep3_scripted_effects.txt
@@ -1431,12 +1431,12 @@ purchase_land_interaction_effect = {
 	}
 	# Purchase land specific
 	scope:actor = {
-		show_as_tooltip = {
+		#show_as_tooltip = { #Unop: We based all script review on the effect taking gold, so comment this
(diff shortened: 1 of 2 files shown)


common/character_interactions/06_ep3_laamp_interactions.txt
```diff
@@ -358,11 +358,6 @@ purchase_land_interaction = {
 			scope:actor = { use_hook = scope:recipient }
 		}
 		scope:actor = { trigger_event = ep3_laamps.0100 }
-		hidden_effect = { # Unop: Give the actor the gold that was taken from the cost = {} block, since it will be taken inside purchase_land_interaction_effect to be given to the recipient
-			scope:actor = {
-				add_gold = scope:purchase_land_cost
-			}
-		}
```
pharaox: This will make the actor to pay the cost, but the recipient will not get anything, correct? Shouldn't we remove the `show_as_tooltip` from the effect instead?

pharaox (changes requested):

PR #369, opened 2026-01-15
All can_create_maa where not replaced (I guess it was updated and I m…

ProZeratul:
…issed it while making the compatible version)


pharaox (approved):
lgtm

PR #379, opened 2026-01-27
Dev/fix event ancestor

### LINKED ISSUE #357: tgp_decision_events.0200 > Ancestors list computation logic issue
From steam comment:
> Great grandparent is not a valid target for reincarnation if grandparent is still alive
> for tgp_decision_events.0200
> grandparents_and_great_grandparents list is first built by putting in dead grandparent
> then as second step, it adds the parents of the ones already in the list
> 
> the logical flaw is that if the grandparent is still alive, they will not be in the list, and their parent (i.e. great grand parent) who is dead will not be considered too

ProZeratul:
Fix #357

@pharaox I've some doubt on the modifier:

```
modifier = {
    is_ruler = yes
    add = 4
}
```

Can dead char still return true for this ? if not this might need some tweaks to see if the GP/GGP was a ruler

### DIFF
--- a/events/decisions_events/tgp_decision_events.txt
+++ b/events/decisions_events/tgp_decision_events.txt
@@ -1042,7 +1042,7 @@ tgp_decision_events.0200 = {
                         add = 2
                     }
                     modifier = {
-                        is_ruler = yes
+                        highest_held_title_tier >= tier_county #Unop: Was is_ruler = yes but since they are dead, it don't work
                         add = 4
                     }
                 }


pharaox (commented):
Looks good, except that we should probably remove the modifier checking `is_ruler` on a dead character.

pharaox:
* **File(s):** `events/decisions_events/tgp_decision_events.txt`

* **Issue:** The Divine Reincarnation event (`tgp_decision_events.0200`) had two bugs: (1) Great-grandparents weren't considered as reincarnation candidates if their children (grandparents) were still alive, due to the two-step list building logic only including dead grandparents initially; (2) The `is_ruler = yes` modifier didn't work for dead characters, causing rulers to not receive proper weighting.

* **Fix:** The fix addresses both issues by: (1) Removing the `is_alive = no` filter when initially building the grandparents list, allowing alive grandparents to be included so their dead parents can be added in step 2, then filtering to only dead characters when selecting the final ancestor; (2) Replacing `is_ruler = yes` with `highest_held_title_tier >= tier_county` which works correctly for dead characters. Additionally fixes an unrelated vanilla bug where multiple desc lines in triggered_desc only showed the first one, and optimizes the family member iterator.

* **Analysis:** The fixes correctly address the reported issue and discovered bug. The approach of including alive grandparents temporarily to capture their dead parents, then filtering at selection time, is sound. Replacing `is_ruler` with `highest_held_title_tier` is the proper solution for dead characters. The desc fix properly uses nested blocks to show multiple descriptions.

_-- Review by Claude Code_

PR #386, opened 2026-02-02
Remove the herd limit before removing nomadic title to a landless nomad

ProZeratul:


pharaox (commented):
@ProZeratul I am using Claude Code to review PRs in this project. The following comment was posted by it. Since the PR has no description. I find it helpful to see this additional context here, let me know if you have any objections.

---
The fix makes sense. A landless nomad with a large herd is a contradiction — they have no land for herders, no way to grow, and get stuck indefinitely. The vanilla threshold of `domicile.herd < 500` was likely meant to let "wealthy" nomads persist, but in practice it creates permanently stuck characters.

The cleanup already handles the transition gracefully — moves the character and courtiers to the domicile location, destroys nomadic titles, and has a chance to create a landless adventurer (if Roads to Power is active). The other conditions in the limit block (`is_at_war = no`, `is_migrating = no`, no active activity) provide sufficient gating to prevent premature cleanup.

PR #392, opened 2026-02-09
Adventurer treasure hunt fix

ProZeratul:
Looking at people reporting treasure contract sending them to europe while being in asia; looks like it come from 1025_set_contract_location_effect, since it as not been updated for the new geographical_region it fallsback on region = world_asia_minor #Something something Bosnia

This is from Pep another modder who I exchange a lot who have done the fix when I pointed at the issue.

I'm not a better creator as he is for text so I probably won't change the strings, so if you have any suggestion @pharaox

### DIFF
--- a/common/on_action/relations/relation_on_actions.txt
+++ b/common/on_action/relations/relation_on_actions.txt
@@ -142,6 +142,7 @@ on_set_relation_lover = {
 						exists = house
 						exists = root.house
 						house = root.house
+						root != this #Unop: In case you fall in love with your primary spouse...
 					}
 				}
 			}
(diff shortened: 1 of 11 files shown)


pharaox:
* **File(s):** `events/dlc/ep3/ep3_laamp_decision_events.txt`, localization files
* **Issue:** Adventurer treasure hunt contracts send players to the wrong regions when in Asia. The vanilla code wasn't updated for the three new geographical regions added in the "All Under Heaven" DLC (`world_asia_east`, `world_asia_north_east`, `world_asia_southeast`), causing treasure hunts to incorrectly fall back to Europe (Bosnia/Asia Minor).
* **Fix:** Adds support for the three new Asian regions in five locations: events 1000/1001, `1025_set_contract_location_effect`, `1025_option_picked_spawn_contract_effect`, and `1025_can_create_legendary_contract_trigger`. Also adds 12 new localization keys for flavor text.
* **Analysis:** The fix is comprehensive and follows vanilla patterns. The additions are placed in logical locations and match the existing structure.

_-- Review by Claude Code_

PR #406, opened 2026-02-26
Fix ep3_contract_event.0007 buy artifact option

### LINKED ISSUE #404: ep3_contract_events assigns artifact to wrong owner
as an adventurer on a treasure hunt contract, should you end up in a city,
pick the first option to haggle(merchant then generates a 2nd artifact)
and pick "buy everything for yourself"
the game will happily take your gold and give the artifact to the merchant.

the error lies on line 1960 from events\dlc\ep3\ep3_contract_events:
set_owner = scope:local_1
local_1 is the generated merchant. it should be:
set_owner = root
which would assign the artifact to the player.

p.s. the way the game decides to convert pedestal artifacts to trinkets/regalia for adventurers is also kinda weird in that it ends up generating an extra "received artifact" notifications and extra events in artifact history, but we probably have bigger fish to fry.

ProZeratul:
This will fix https://github.com/ProZeratul/CrusaderKings3UnofficialPatch/issues/404

### DIFF
--- a/events/dlc/ep3/ep3_contract_events.txt
+++ b/events/dlc/ep3/ep3_contract_events.txt
@@ -1955,10 +1955,10 @@ ep3_contract_event.0007 = {
 			target = scope:local_1
 			gold = major_gold_value
 		}
-		change_treasure_type = yes
-		scope:newly_created_artifact = {
-			set_owner = scope:local_1
-		}
+		#change_treasure_type = yes  #Unop: Disable this & below since it's already made in the immediate part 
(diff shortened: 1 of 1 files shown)


pharaox (changes requested):

ProZeratul:
Then it's even worse, the artifact seems to be created in ep3_contract_event.0006 and another is created in ep3_contract_event.0007 using the same scope name but representing 2 different artifacts. I'll leave this one open and will check in-game when I get ~~my computer back~~ some spare time

PR #409, opened 2026-02-27
Add missing Preservation tenet checks to Monasteries 2-8

### LINKED ISSUE #410: Preservation tenet only enables 1st level of monastery
https://forum.paradoxplaza.com/forum/threads/cant-upgrade-monasteries-with-preservation.1886059/

The hindu tenet preservation, per description, is supposed to enable construction of the entire line of monasteries, but only checks for the tenet for monastic_schools_01

I've created a pull request that adds checks to monastic_schools_02 through monastic_schools_08
But the tooltip is not localized properly:

Image

andathree:
This should fix
https://forum.paradoxplaza.com/forum/threads/cant-upgrade-monasteries-with-preservation.1886059/

The tooltip is a little wonky though.
Screenshot 2026-02-26 215242

ideally it would say "Your Faith has the Preservation tenet"; not sure if that's already in localization 

Fixes #410

### DIFF
--- a/common/buildings/00_temple_buildings.txt
+++ b/common/buildings/00_temple_buildings.txt
@@ -1,4 +1,4 @@
-﻿@illustration_christian = "gfx/interface/illustrations/holding_types/temple_christian.dds"
+@illustration_christian = "gfx/interface/illustrations/holding_types/temple_christian.dds"
 @illustration_generic_grass = "gfx/interface/illustrations/holding_types/temple_generic_grass.dds"
 @illustration_generic_arid = "gfx/interface/illustrations/holding_types/temple_generic_arid.dds"
 @illustration_india = "gfx/interface/illustrations/holding_types/temple_india.dds"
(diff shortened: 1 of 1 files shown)


pharaox:
* **File(s):** `common/buildings/00_temple_buildings.txt`
* **Issue:** Vanilla `monastic_schools_01` already includes an `OR` check allowing the Preservation tenet (`monastic_schools_allowed_in_every_holding_type`) to bypass the temple building requirement, but tiers 2–8 are missing this check, preventing players with the Preservation tenet from upgrading monasteries in non-temple holdings.
* **Fix:** Wraps each `has_building_or_higher = temple_XX` check in an `OR` block adding the Preservation tenet as an alternative. For tiers with a `temple_02`+ requirement, also requires the equivalent holding-level building via `building_requirement_castle_city_church = { LEVEL = XX }`, correctly substituting the temple tier for castle/city holdings.
* **Analysis:** The logic is correct and consistent with the existing vanilla pattern. However, each changed `has_building_or_higher` line is commented out while still being re-added inside the `OR` block below — the line is not removed, so commenting it out is misleading and inconsistent with the established pattern in `monastic_schools_01`. The fix should simply wrap the original line in an `OR` block without commenting it out.
* **Recommendations:** Remove the `# has_building_or_higher = ...` commented lines (13 in total) since the same check is preserved inside the `OR` blocks.

_-- Review by Claude Code_

@andathree Thanks for the fix! Could you follow the recommendation above and remove the commented out lines? This is purely a style we try to be consistent with in Unop.

pharaox (approved):

PR #434, opened 2026-05-02
Fix missing description in chariot_race.4030 seduce option toast

### LINKED ISSUE #424: Missing description in seduce event in a chariot race
# Summary
When trying to seduce another character through a chariot race event, when the 1st option is chosen, the pop-up that happens after 1 in-game day is missing a description.

# Steps to Reproduce
Choose the "Seduce" intent while hosting a "Chariot Race" and choose a character, if the event shown in the screenshot happens throughout the race, pick the option also shown in the screenshot (as pointed by the blue arrow that I edited in). After that, the pop-up that happens 1 day later will be missing description (as also shown in the other screenshot).

# Expected Result
The pop-up has a description that relates to the event's outcome.

# Actual Result
There is no description.

pharaox:
Fixes #424

The `send_interface_toast` for `chariot_race.4030` option B (seduce) success and failure used `type = event_generic_good_text` / `event_generic_bad_text`, which require a `desc =` (rendered as `$event_message_text$`). Since neither toast supplied a `desc`, the engine printed the literal `DESCRIPTION` placeholder in the side notification.

The fix comments out the `type =` lines so the toasts default to a non-`_text` style (matching option A in the same event, which is unaffected by the bug).

### DIFF
--- a/events/activities/chariot_race_activity/chariot_race_ongoing_events.txt
+++ b/events/activities/chariot_race_activity/chariot_race_ongoing_events.txt
@@ -3257,7 +3257,8 @@ chariot_race.4030 = {
 				}
 				desc = chariot_race.4030.b.success
 				send_interface_toast = {
-					type = event_generic_good_text
+					#Unop event_generic_good_text requires a desc which is not provided here, leading to a "DESCRIPTION" placeholder
+					#type = event_generic_good_text
 					title = chariot_race.4030.b.success
 					left_icon = root
 					right_icon = scope:john
@@ -3331,7 +3332,8 @@ chariot_race.4030 = {
 				}
 				desc = chariot_race.4030.b.failure
 				send_interface_toast = {
-					type = event_generic_bad_text
+					#Unop event_generic_bad_text requires a desc which is not provided here, leading to a "DESCRIPTION" placeholder
+					#type = event_generic_bad_text
 					title = chariot_race.4030.b.failure
 					left_icon = root
 					right_icon = scope:john


pharaox:
With the proposed fix:

events/activities/chariot_race_activity/chariot_race_ongoing_events.txt
```diff
@@ -3257,7 +3257,8 @@ chariot_race.4030 = {
 				}
 				desc = chariot_race.4030.b.success
 				send_interface_toast = {
-					type = event_generic_good_text
+					#Unop event_generic_good_text requires a desc which is not provided here, leading to a "DESCRIPTION" placeholder
+					#type = event_generic_good_text
```
ProZeratul, line 3261: Why not changing it to event_generic_good?

events/activities/chariot_race_activity/chariot_race_ongoing_events.txt
```diff
@@ -3331,7 +3332,8 @@ chariot_race.4030 = {
 				}
 				desc = chariot_race.4030.b.failure
 				send_interface_toast = {
-					type = event_generic_bad_text
+					#Unop event_generic_bad_text requires a desc which is not provided here, leading to a "DESCRIPTION" placeholder
+					#type = event_generic_bad_text
```
ProZeratul, line 3336: Why not changing it to event_generic_bad ?

ProZeratul (approved):
There are type variants without the text part that is causing the issue (since the DESC is missing and it would be better to replace to use them instead of removing the type completly IMHO)

events/activities/chariot_race_activity/chariot_race_ongoing_events.txt
pharaox: We could, but removing it is more consistent with the rest of the code. There are 22 occurrences of `send_interface_toast`, and only these 2 have `type`. And adding a type converts the "toast" into a "message".
It is not really clear which of the 2 options better reflects the original intent, but I tend to prefer removing the type. Let me know if you disagree and I will change it.

events/activities/chariot_race_activity/chariot_race_ongoing_events.txt
pharaox: See above.

events/activities/chariot_race_activity/chariot_race_ongoing_events.txt
ProZeratul: Ok I didn't check within the file itself as indeed other events use them without type, So I'm fine with removing it then

events/activities/chariot_race_activity/chariot_race_ongoing_events.txt
ProZeratul: Same thing as above, so go with this !

PR #435, opened 2026-05-02
Fix missing NOT_HOF_INTERACTION_HAS_HOUSE_MODIFIER localization

pharaox:
Fixes #421

`ep3_hof_ask_for_influence_interaction` ("Ask for Clerical Influence") in `common/character_interactions/06_ep3_interactions.txt` has a `custom_description` whose inner trigger contains `NOT = { has_house_modifier = ep3_requested_faith_support_modifier }`. When the actor's house already has the modifier, the engine renders the failure tooltip with the `NOT_` prefix — but vanilla only defines `HOF_INTERACTION_HAS_HOUSE_MODIFIER`, so the raw key name is shown.

Adds `NOT_HOF_INTERACTION_HAS_HOUSE_MODIFIER` aliased to the existing positive-form text via `$HOF_INTERACTION_HAS_HOUSE_MODIFIER$`, so the two stay in sync.


pharaox:
With the proposed fix:

PR #438, opened 2026-05-02
Allow transport contract targets across water

### LINKED ISSUE #418: Transport Contract Targets can only be random_neighboring_top_liege_realm_owner
task_contracts\laamp_transport_contracts.txt
When generating transport contracts the game only uses random_neighboring_top_liege_realm_owner to select the target. The issue isn't apparent in europe/africa, but adventurers in the middle of any large empire/hegemony will be offered transport contracts that are 9 months or more away. also, random_neighboring_top_liege_realm_owner doesn't include water. The worst-case scenario is 867 japan, where the only valid target is the enishi ruler north of japan, because japan doesn't have any other land neighbors.

To play devils advocate, maybe travel within a realm is safe enough for rulers that they don't see a need to offer contracts. But pay isn't scaled to distance, so a I think a better choice is to use random_neighboring_and_across_water_realm_owner

random_neighboring_and_across_water_realm_same_rank_owner could be another alternative, so counts talk to counts, dukes talk to dukes, kings talk to kings, emperors talk to emperors, and the hegemon talks to nobody(he's too busy receiving tributes anyways)

pharaox:
Fixes #418

All six `laamp_transport_*` contracts (`vip`, `artifact`, `gold`, `ward`, `explorer`, `animal`) picked targets via `random_neighboring_top_liege_realm_owner` (and gated creation with `any_neighboring_top_liege_realm_owner`), which only walks land neighbors. Rulers in landlocked or water-locked positions — notably 867 Japan — end up with no realistic target, or get sent on contracts that take 9+ months one way.

Switch both the `valid_to_create` gate and the `on_create` random pick to the `_and_across_water` variant in each of the six contracts so adventurers across narrow seas count as valid targets.

The reporter's named alternative (`random_neighboring_and_across_water_realm_owner`) doesn't exist in vanilla; the matching member of the file's existing iterator family is `random_neighboring_and_across_water_top_liege_realm_owner`. The same-rank variant the reporter also suggested would be a balance change rather than a fix and is out of scope here.

`make tiger` reports no new errors or warnings introduced by this change.

### DIFF
--- a/common/task_contracts/laamp_transport_contracts.txt
+++ b/common/task_contracts/laamp_transport_contracts.txt
@@ -1,0 +1,2029 @@
+﻿##################################################
+# INFO
+# Transport contracts for laamps. Any contract that require usage of the travel system to complete, i.e. fetch quests, escort quests, etc.
+#
+##################################################
+
+
+
+##################################################
+# GROUPS (for populate_task_contracts_for_area)
+#
+# laamp_contracts_transport_group
+#
+#
+#
+
+
+##################################################
+#
+# laamp_transport_vip - Escort Emissary - Escort a character's Courtier to another character
+# laamp_transport_artifact - Artifact Delivery - Transfer an artifact from one character to another
+# laamp_transport_gold - A Golden Gift - Transfer gold from one character to another
+# laamp_transport_ward - Ward Transfer - Escorting a Ward to a Guardian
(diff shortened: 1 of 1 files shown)


pharaox:
Accidentally, I came across this independently from the reporter and actually experimented with their fix and other changes. I can confirm adding `_and_across_water` improves the user experience with such contracts, especially in certain regions. Other improvements are also possible, but this is the most straightforward / less invasive one.

PR #439, opened 2026-05-02
Fix major-decision culture rewards for divergent/hybrid cultures

### LINKED ISSUE #417: Major decisions that modify cultures weren't designed with divergent/hybrid cultures in mind
scripted_effects\00_major_decisions_scripted_effects
form_carolingian_empire_scripted_effect
Restore Carolingian Borders can be taken with any Frankish heritage culture, but only grants innovations to the French. (And to add insult to injury, you'd have to hybridize twice afterwards to get them, since same heritage cultures can't hybridize)
Compare to Unify Italy, where the rewards are given to all Latin heritage cultures.

scripted_effects\10_dlc_tgp_japan_scripted_effects
japanese_become_shogun_reward_effect
Formalize Shogunate has no culture requirement. If the decision taker's culture has Fragile Peace, it is replaced with Bushido, else Bushido is unlocked for culture reform. Rulers who culture convert/diverge cultures/hybridize before taking the decision will likely end up with a pack of Japanese vassals stuck in Fragile Peace. I think it would be fair to run the replace-or-unlock process for japonic heritage cultures as well.

pharaox:
Fixes #417

Two related vanilla bugs in major-decision scripted effects that hardcode a single culture instead of iterating the heritage pillar the decision is gated on.

### Bug A — `form_carolingian_empire_scripted_effect`
`reform_carolingian_empire_decision` is gated by `culture = { has_cultural_pillar = heritage_frankish }`, so any Frankish-heritage culture (Occitan, German, Dutch, Flemish, etc.) can take it — but the effect grants `innovation_knighthood`, `innovation_royal_prerogative`, and `innovation_heraldry` only to `culture:french`.

Mirror the Italian unification pattern (`roman_restoration.0140`): iterate `every_culture_global` filtered by `has_cultural_pillar = heritage_frankish`.

### Bug B — `japanese_become_shogun_reward_effect`
Formalize Shogunate has no culture restriction on the decision, but the Imperial Peace → Bushido swap (or Bushido unlock) only ran on the decision taker's own culture. A ruler who hybridized/diverged/converted before taking the decision left their Japanese vassals stuck on Fragile Peace forever, since the swap never reached the other Japonic-heritage cultures in the realm.

Replace the inner `culture = { ... }` scope with `every_culture_global = { limit = { has_cultural_pillar = heritage_japonic } ... }` doing the same swap-or-tooltip work.

### Incidental
The new vanilla copy of `10_dlc_tgp_japan_scripted_effects.txt` surfaced a tiger error: `tgp_dynasty_house_name_setup_effect` references `dynasty:korean_wang_haeju`, which is commented out in vanilla `common/dynasties/05_tgp_dynasties.txt`. The call uses `?=` so it's a silent no-op in-game; commented out here too to keep tiger clean.

`make tiger` reports no new errors or warnings introduced by this change.

### DIFF
--- a/common/scripted_effects/00_major_decisions_scripted_effects.txt
+++ b/common/scripted_effects/00_major_decisions_scripted_effects.txt
@@ -1410,7 +1410,9 @@ form_carolingian_empire_scripted_effect = {
 		}
 	}
 	# Innovations
-	culture:french = {
+	#culture:french = { #Unop Reward all Frankish-heritage cultures, not just French
+	every_culture_global = {
+		limit = { has_cultural_pillar = heritage_frankish }
 		if = {
 			limit = {
 				NOT = { has_innovation = innovation_knighthood }
--- a/common/scripted_effects/00_major_decisions_scripted_effects.txt
+++ b/common/scripted_effects/00_major_decisions_scripted_effects.txt
@@ -1,0 +1,4209 @@
+﻿create_roman_empire_scripted_effect = {
+	primary_title = { save_scope_as = old_primary_title }
+	add_trait = augustus
+	if = {
+		limit = { has_dlc_feature = legends }
+		create_legend_seed = {
+			type = legitimizing
+			quality = illustrious
+			chronicle = new_title
+			properties = {
+				title = title:h_roman_empire
+				founder = root
+			}
+		}
+	}
+	#Create Roman Empire, shift dejure/history/laws, destroy e_byzantium.
+
+	if = {
+		limit = { capital_county = { this = title:c_byzantion } }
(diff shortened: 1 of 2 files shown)


pharaox:
@andathree Can you confirm if this matches your expectations before we merge this?

ProZeratul:
I remember seeing form_carolingian_empire_scripted_effect only affecting french so I'm good with this.
I've no objection to the open the effect to all japonic heritage for Bug 2

So overall I've nothing to say @pharaox merge as you see fit (as it's a draft)

pharaox:
@ProZeratul Thanks! Merging then.

PR #448, opened 2026-05-02
Fix Empire of Germania creation blocked for Roman Emperors after HRE restoration

### LINKED ISSUE #440: Empire of Germania creation blocks Roman Emperor after HRE restoration
**Reported on Steam by `Nider001` on `2026-03-10 14:14`** ([source](https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329)).

> I'd like to make a report while at it. Empire of Germania has various restrictions on who can create it, to encourage restoring HRE instead. In my current game as a HRE emperor I went a step further and restored the Roman Empire, which destroyed my HRE title. At this point Empire of Germania came back and absorbed most titles I de jure drifted into HRE. As a result, I am now stuck with this large empire I can't create, and the Roman Emperor is not allowed to restore HRE either (is_roman_emperor_trigger is being checked by the Restore HRE decision):  https://imgur.com/a/3oY0UGV
>
> I suggest relaxing the restrictions on Empire of Germania title creation, including allowing any rulers who fulfill the "is_roman_emperor_trigger = yes" condition to do so, since they aren't allowed to restore HRE.

Empire of Germania creation tooltip showing the block

pharaox:
Fixes #440

After a Catholic HRE Emperor takes `restore_roman_empire_holy_decision`, the HRE title is destroyed but `e_germany` is gated to non-Catholic faiths and `restore_holy_roman_empire_decision` is permanently blocked by the `flag_restored_roman_empire` global, leaving the central-European realm without any usable empire title.

This PR releases the Catholic gate on `e_germany.can_create` when HRE refoundation is no longer reachable (i.e. `flag_restored_roman_empire` is set). Catholics on any other path still see the original "reform the HRE instead" gate.

## Commits

1. Add base game file `common/landed_titles/00_landed_titles.txt`
2. Apply the fix; add a `ck3-tiger.conf` ignore rule for two preexisting harmless duplicate-field warnings in the newly added vanilla file (`can_use_nomadic_naming` and a duplicate `province = 5505`).


pharaox:
With the fix, I am able to create the empire of Germania after restoring the Roman Empire as the HRE emperor:

ProZeratul (approved):

Issue #284, opened 2025-09-05
Multiple strings missing localization


anuicswtor42:
<img width="2560" height="1080" alt="Image" src="https://github.com/user-attachments/assets/758e6e1c-eac4-4571-bf10-8a59ee251599" />
<img width="2560" height="1080" alt="Image" src="https://github.com/user-attachments/assets/c9949436-d92e-4a7d-9bd8-93148df839d0" />
<img width="2560" height="1080" alt="Image" src="https://github.com/user-attachments/assets/abc8139b-c5ff-4b93-9376-44839fb1d95d" />
<img width="2560" height="1080" alt="Image" src="https://github.com/user-attachments/assets/908cef5a-f215-4c26-bbbd-5f2d8d03d6a0" />
<img width="2560" height="1080" alt="Image" src="https://github.com/user-attachments/assets/aaa3d68a-fd4a-46b9-86c0-301f5785c137" />
<img width="2560" height="1080" alt="Image" src="https://github.com/user-attachments/assets/6e1ab0bb-2991-4b65-b035-c25fede943c8" />
<img width="2560" height="1080" alt="Image" src="https://github.com/user-attachments/assets/bf9665c5-e4b2-480f-803e-a27dd5932679" />
<img width="2560" height="1080" alt="Image" src="https://github.com/user-attachments/assets/c2a29675-adf2-4be2-84f0-dd581137e55f" />
<img width="2560" height="1080" alt="Image" src="https://github.com/user-attachments/assets/25ef69cd-4805-41a8-97a3-8e7238a268ac" />

anuicswtor42:
another string that I've found that is missing localization 

<img width="2560" height="1080" alt="Image" src="https://github.com/user-attachments/assets/0595a96e-1bfc-4c37-a4d3-79896bae40ee" />

anuicswtor42:
this last one is as an adventurer, not a landed character

ProZeratul:
Most important ones are now in 8.0.38, the debug ones will be added in the next

anuicswtor42:
<img width="2560" height="1080" alt="Image" src="https://github.com/user-attachments/assets/6933b941-b178-4e52-80e6-19f5dd2bccd4" />

anuicswtor42:
New string missing loc.

anuicswtor42:
New string missing localization in debug mode - bottom right of the screen

anuicswtor42:
[...]

PR #453, opened 2026-05-03
Add missing action_join_tributary_warcombined_label loc key

pharaox:
Refs #284 (#3477963837)

Adds the missing `action_join_tributary_warcombined_label` loc key, which was rendered raw in the "Current Situation" panel under "You can Join Wars".

Vanilla defines `action_join_tributary_war_combined_label` (with underscore between `war` and `combined`), but the engine asks for the underscore-stripped form. Unop already covers the matching `_group_label` variant in the same way; this commit adds the row-level companion.


pharaox:
Closing — the underlying engine bug is fixed in 1.19. Verified locally: the "You can Join Wars" row in the Current Situation panel renders correctly under 1.19, so the engine is asking for the documented `_combined_label` form again. The reporter's screenshots in #284 were on 1.18.

The existing Unop redirect for `action_join_tributary_warcombined_group_label` is now also a no-op on 1.19 and could be cleaned up in a follow-up if desired.

PR #454, opened 2026-05-03
Fix adopted bastard losing house when secret is exposed

### LINKED ISSUE #213: An adopted bastard can change their house / dynasty if their secret gets exposed
From Steam comments:

> Emperor2000 6 Jul @ 12:50am 
> Adopting a Bastard, while their Parent has the Secret of it can result in losing the Child and/or the adopted Child changes the House/Dynasty, when the Secret gets exposed, because the Secret is not cleared from the old Parent.

Indeed, removing secrets from real parents is completely missing from `adopt_effect` and `adventurer_adopt_effect`, and should be perhaps added.

pharaox:
Fixes #213

When a bastard child is adopted, the secret about their true heritage (`secret_unmarried_illegitimate_child` or `secret_disputed_heritage`) remains attached to the bio parents. When that secret is later exposed, vanilla re-binds the child to the bio father (changing house/dynasty) or adds the `disputed_heritage` trait — silently undoing the adoption.

This adds a persistent `unop_was_adopted` character flag in `adopt_effect` and `adventurer_adopt_effect`, then skips the child-side consequences in both `secret_exposed_owner_effects_effect` and `secret_unmarried_illegitimate_child.on_expose` (vanilla has duplicate logic in both places) when the flag is present. Parent-side consequences (adultery traits, opinions) still fire, so secret owners keep their leverage.

Also fixes four pre-existing vanilla scope bugs surfaced by ck3-tiger once the affected files were brought into the mod tree:
- `00_bastard_secrets.txt`: two `secret_exposed_owner_effects_effect` calls in `on_expose` ran in secret scope; wrapped in `scope:secret_owner` to match the convention in `00_secret_types.txt`.
- `00_bastard_secrets.txt`: `add_courtier = root` in the hostage-return path used the secret as the courtier; corrected to `prev` (the child).
- `00_secret_effects.txt`: `save_my_known_father_effect` referenced `scope:father` which is unset at one of its two call sites; corrected to `father` (the implicit relation of the discoverer).

Note: characters adopted before this fix will not have the flag and remain vulnerable on exposure. There is no clean way to retroactively flag them.

### DIFF
--- a/common/secret_types/00_bastard_secrets.txt
+++ b/common/secret_types/00_bastard_secrets.txt
@@ -1,0 +1,456 @@
+﻿
+secret_disputed_heritage = {
+	category = adultery
+
+	is_valid = {
+		secret_disputed_heritage_is_valid_trigger = {
+			TARGET = scope:secret_target
+		}
+	}
+
+	is_shunned = {	
+		bastard_secret_is_shunned = {
+			TARGET = scope:secret_target
+			PARTICIPANT = scope:target
+		}
+	}
+
+	is_criminal = {
+		bastard_secret_is_criminal = {
+			TARGET = scope:secret_target
+			PARTICIPANT = scope:target
+		}
+	}
+
+	on_owner_death = {
+		save_scope_as = secret
+		scope:secret_target = {
+			save_scope_as = child
+		}
+
+		###SET NEW OWNERSHIP###
+		#Should we set new secret owner(s)?
+		if = {
+			limit = {
+				#Child knows about the secret
+				is_known_by = scope:child
+				scope:child = { is_alive = yes }
+			}
+			set_secret_owner = scope:child
+		}
+	}
+
+	on_discover = {
+		save_scope_as = secret
+		scope:secret_target = {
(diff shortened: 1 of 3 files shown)


pharaox:
Verified in-game: after exposing a "Child Heritage" secret against the mother of my adoped house, she and the real father got "Fornicator", but the child remained in my house.

common/secret_types/00_bastard_secrets.txt
```diff
@@ -449,7 +449,7 @@ secret_unmarried_illegitimate_child = {
 			if = {
 				limit = { exists = scope:unhostaged_child }
 				return_hostage = yes
-				scope:warden = { add_courtier = root }
+				scope:warden = { add_courtier = prev } #Unop: root is the secret in on_expose, but add_courtier needs the child character
```
ProZeratul: Since the unhostaged_child scope is registered, I'd prefer to use it than using prev here

pharaox: Changed as suggested

PR #464, opened 2026-05-08
Skip auto-spawning a coronation inspiration when one is already owned

### LINKED ISSUE #442: Tier promotion always generates crown inspiration, blocking regalia-required coronations
**Reported on Steam by `Nider001` on `2026-03-20 03:05`** ([source](https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329)).

> Bug/oversight report: Increasing your title tier to kingdom or empire always generates a crown inspiration in your court to aid with the coronation, even though a good number of rulers uses regalia instead. Even worse if you have mandatory imperial anointment, because it becomes impossible to host a coronation until you commission a regalia artifact (the cheapest activity options that allow hosting a coronation without an artifact are blocked)

pharaox:
Fixes #442

In `court_maintenance.0011` (gained a Royal Court), the call to `create_proper_coronation_artifact_setup` ran unconditionally on the first kingdom-tier promotion, even if the character already had a suitable crown or regalia artifact. `realm_maintenance.2001` and `ach_yearly_events.1005` already gate the same call on `coronation_has_proper_artifact_trigger = no`; mirror that here.

### DIFF
--- a/events/court_maintenance_events.txt
+++ b/events/court_maintenance_events.txt
@@ -332,7 +332,11 @@ court_maintenance.0011 = {
 			}
 
 			# Artifact for Coronations setup
-			create_proper_coronation_artifact_setup = yes
+			#Unop Don't spawn an inspiration if a proper coronation artifact is already owned
+			if = {
+				limit = { coronation_has_proper_artifact_trigger = no }
+				create_proper_coronation_artifact_setup = yes
+			}
 		}
 		
 		if = {


pharaox:
With the fix, no artifact:


With the fix, crown artifact already available:

pharaox:
@ProZeratul I think the issue that was found during the investigation is valid and so I am fixing it, but it's not quite the issue that the reporter reported, so I left some questions. We could merge this one if it looks ok, or wait for this additional clarification before merging.

ProZeratul:
@pharaox This can be merge, but indeed re-reading the original issue it seems like they were preventend to setup a coronation despite having a regalia or crown. I've taken a look but didn't find anything obvious for now so with no further information I don't know how to deal with the rest

ProZeratul (approved):

PR #465, opened 2026-05-08
Fix belligerent_levy_contribution_mult using wrong modifier loc

### LINKED ISSUE #443: belligerent_levy_contribution_mult uses wrong modifier loc key
**Reported on Steam by `Nider001` on `2026-03-24 13:57`** ([source](https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329)).

> Localization bug report: modifier string for belligerent levy contribution multiplier bonus incorrectly uses the "base contribution" text instead of the multiplier one. Localization texts for _add and _mult effects are currently identical, but it should be modifier_contribution instead of modifier_base_contribution in the second line.
>
> belligerent_levy_contribution_add:0 "$belligerent_vassal$ [levy|E] [modifier_base_contribution|E]"
> belligerent_levy_contribution_mult: "$belligerent_vassal$ [levy|E] [modifier_base_contribution|E]"

pharaox:
Fixes #443

The English loc for `belligerent_levy_contribution_mult` reused the same `[modifier_base_contribution|E]` text as the `_add` variant, instead of `[modifier_contribution|E]` (compare with the symmetric `belligerent_tax_contribution_add` / `_mult` pair, which is correct).

Adds an Unop override in `unop_fixed_keys_l_english.yml` that swaps the wrong concept reference.


pharaox:
Without the fix:


With the fix:

PR #466, opened 2026-05-10
Add all missing zealot vassal stance check in coronation_events.6130

ProZeratul:
This will prevent ALL vassal to get the opinion duplicate (instead of just the zealous ones)

@pharaox I find this effect fishy, I've not changed it much except the obvious missing zealous check, but I still find it looking odd (as the whole thing looks duplicated) but I don't really know what would be the correct way to clean this. What do you think ?

### DIFF
--- a/events/activities/coronation_activity/coronation_events_6.txt
+++ b/events/activities/coronation_activity/coronation_events_6.txt
@@ -12145,6 +12145,7 @@ coronation_events.6130 = {
 							faith.fervor >= 100
 						}
 						every_vassal = {
+							vassal_stance = zealot #Unop: Missing zealot vassal stance check
 							custom = custom.every_zealot_vassal
 							add_opinion = {
 								target = scope:host
@@ -12167,6 +12168,7 @@ coronation_events.6130 = {
 							faith.fervor >= 80
 						}
 						every_vassal = {
+							vassal_stance = zealot #Unop: Missing zealot vassal stance check
 							custom = custom.every_zealot_vassal
 							add_opinion = {
 								target = scope:host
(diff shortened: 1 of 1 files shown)


pharaox (commented):
* **File(s):** `events/activities/coronation_activity/coronation_events_6.txt`
* **Issue:** In `coronation_events.6130`, the anointment-specific reward chain (line 12132 onward, gated on `coronation_type_anointment`) iterates `every_vassal` with the desc `custom.every_zealot_vassal` but applies `holy_anointment_opinion` to all vassals, not just zealots. The non-anointment chain above uses proper `limit = { faith = ... }` filters, but the zealot chain is missing the corresponding stance filter in all six fervor tiers (>=100, >=80, >=60, >=40, >=20, else).
* **Fix:** Adds `vassal_stance = zealot` as an iterator filter to each of the six `every_vassal` blocks, so opinions only apply to zealot vassals. The direct-on-iterator syntax matches existing Unop usage (e.g. `common/buildings/00_temple_buildings.txt`).
* **Analysis:** Minimal, correct, and complete — all six tiers are covered and the change matches the `custom.every_zealot_vassal` desc already present.

> I find this effect fishy ... the whole thing looks duplicated

Not really duplicated — they're two separate reward chains: the upper one (~11923–12130) is the general/non-anointment branch (same-faith / different-faith vassals, tiers `>=90/80/70/60` and `<30/20/10`), the lower one (12132–12272) is anointment-only (`this = root` + `coronation_type_anointment`), targets *zealot* vassals, uses different fervor tiers (`>=100/80/60/40/20/else`), and also adds legitimacy and piety. Parallel by shape, different by intent. The missing `vassal_stance = zealot` filter is the real bug; the rest is just verbose, not wrong. I'd merge as-is.

pharaox (approved):

PR #467, opened 2026-05-10
Update bookmark_events.txt

ProZeratul:
In 867 play as the recommended character in China Huang Chao, let some day pass until the event "The Withering of Righteousness" start , notice that both options to start a war doesn't work

### DIFF
--- a/events/bookmark_events.txt
+++ b/events/bookmark_events.txt
@@ -2387,6 +2387,7 @@ bookmark.0500 = {
 		start_war = {
 			casus_belli = admin_barbarian_conquest_cb
 			target = title:d_biansong.holder
+			target_title = title:d_biansong #Unop: Without this, the war won't start & error will spawn in logs
 		}
 
 		spawn_army = {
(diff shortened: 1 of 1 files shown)


pharaox (approved):
* **File(s):** `events/bookmark_events.txt`
* **Issue:** In the 867 bookmark event `bookmark.0500` ("The Withering of Righteousness", Huang Chao), options B and D each declare `start_war` with `admin_barbarian_conquest_cb` against `title:d_biansong.holder` but omit `target_title`. The CB is defined with `target_titles = neighbor_land_or_water` and `target_title_tier = duchy`, so a target title is required — without it the war silently fails and an error is logged.
* **Fix:** Adds `target_title = title:d_biansong` to both `start_war` blocks, matching the `target = title:d_biansong.holder` already present and consistent with other `start_war` usages in the same file (e.g. line 1506).
* **Analysis:** Minimal, correct, and complete — these two spots are the only event-side uses of `admin_barbarian_conquest_cb` in the codebase. Comment style and indentation match the surrounding code.

PR #471, opened 2026-05-15
Fix is_landless_adventurer

pharaox:
Fix `is_landless_adventurer` to check if the character has a domicile. Right now, it only checks the government, but this returns true also for landless adventurer **followers** (government checks return true on courtiers for their liege's government).

The main issue is that this trigger is being used in `is_playable_character`, which implies that the character is a ruler (rulers are non-playable). Other triggers included in `is_playable_character` already exclude non-rulers in some way - by checking `has_domicile` and `is_house_head`. Only the first one is appropriate with landless adventurers.

The negative effect is hard to estimate but it's likely big. There are > 700 occurrences of `is_playable_character` in the codebase and landless adventurer followers are currently passing the filter. This trigger is also heavily used by mods. I came across the issue when trying to complete RICE activities as a LAAMP where guests were filtered by `is_playable_character = no` - suddently I couldn't invite anyone and therefore even start any of these activities. This worked before (e.g. in 1.17), I contributed the RICE changes to make it work personally.

I checked quickly other usages of `is_landless_adventurer` (52 in vanilla) for **intended** use on courtiers but couldn't find any, so I think it's safe to merge the fix as proposed.

### DIFF
--- a/common/scripted_triggers/00_available_for_events_triggers.txt
+++ b/common/scripted_triggers/00_available_for_events_triggers.txt
@@ -869,6 +869,7 @@ is_landless_nomad = {
 # This trigger checks if a character is a landless adventurer
 is_landless_adventurer = {
 	government_has_flag = government_is_landless_adventurer
+	has_domicile = yes #Unop Exclude landless adventurer followers
 }
 
 # This trigger checks if a character is a governor in an admin realm (implying they are landed)


pharaox (commented):
### Fix is_landless_adventurer

* **File(s):** `common/scripted_triggers/00_available_for_events_triggers.txt`
* **Issue:** `is_landless_adventurer` only checked `government_has_flag`, which returns true for courtiers/vassals of a LAAMP liege. Landless adventurer *followers* therefore passed `is_playable_character` (700+ usages), incorrectly treated as rulers — e.g. breaking guest filtering for RICE activities.
* **Fix:** Add `has_domicile = yes` so only the adventurer leader (the domicile holder) matches.
* **Analysis:** Correct and minimal. Consistent with sibling triggers (`is_landless_nomad`, `is_landless_administrative`, `is_landless_soryo`), which all add discriminators on top of the government check; `is_landless_adventurer` was the lone outlier. A full sweep of all 98 positive vanilla usages found none that intentionally rely on followers matching — they all run on a leader/host/liege/independent scope, and the follower-aware code paths use `government_has_flag` directly rather than this trigger. Tiger passes clean.
* **Recommendations:** None on the code. Since this changes playable-character filtering broadly, an observer-mode + `error.log` check focused on LAAMP camps is worthwhile (the RICE/LAAMP repro in the description already serves as functional verification).

ProZeratul:
I wouldn't have changed it like that, I'd have changed this inside the **is_playable_character** like what was done to exclude barony tier from landed characters. (just to be extra cautious)

But this seems indeed to be check everywhere else, just have to keep in mind this one might be a potential issue in the future if pdx change it in the future

ProZeratul (approved):

PR #472, opened 2026-05-16
Fix pilgrimage.2004 and pilgrimage.6001 poaching another ruler's spouse/heir/courtiers

### LINKED ISSUE #450: pilgrimage.6001 lets travelling rulers steal spouse, children, and courtiers
**Reported on Steam by `Emperor2000` on `2026-04-26 10:52`** ([source](https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329)).

> This might be an big Request, but could this Mod please fix Events, where travelling Rulers can steal your Spouse, Children and employed Courtiers?

**Follow-up by `Emperor2000` on `2026-05-03 04:59`:**

> @pharaox Now I have the ID, who can steal the Spouse of an Ruler, the ID is pilgrimage.6001.

> Another one for potential Friends is pilgrimage.2004, no checks for Spouses, Child of Ruler, Court Position, Councillor or Accolade.

pharaox:
Fixes #450

`pilgrimage.2004` and `pilgrimage.6001` add the friend to the player's court via `add_to_entourage_court_and_activity_effect` (which calls `add_courtier`), guarded only by `is_ruler = no`. A non-ruler who is another ruler's spouse, child, councillor, knight, or court-position holder is still pulled out of their court.

The court-add is incidental in both event chains — nothing downstream depends on the friend being a courtier of `root`, so the friendship and opinion can still form without the poach.

**Fix 1:** Gate the court-add with the vanilla `can_recruit_character_to_court_trigger` (the canonical recruit gate, already used inside event options elsewhere, e.g. `travel_danger_events_arky.txt`, `feast_tsagaan_sar_events.txt`).

- `pilgrimage.6001`: add the trigger to the existing `if` limit around `add_to_entourage`.
- `pilgrimage.2004`: revert the prior `is_ruler = no` in the selection trigger `pilgrimage_2004_potential_friend_trigger` (it was added to silence the "Trying to put character in court despite being a ruler" script error) and instead wrap `add_to_entourage` in an `if` with `is_ruler = no` + the trigger, mirroring `pilgrimage.6001`. This keeps the friendship forming, avoids suppressing the event, and still avoids the original script error since `add_courtier` no longer runs on a ruler.

**Fix 2:** `can_recruit_character_to_court_trigger`, see the comment on the issue.

Wrapped the checks in `OR`, mirroring the two adjacent spouse-employment blocks (`OR = { #Is employed in some way }`), which is the intended structure. This is the canonical recruit trigger, so the fix also corrects **Invite to Court**, **Noble Family Adoption**, accolade recruitment, and several other events, where employed courtiers of other rulers could be recruited. See #450 for the full analysis.

### DIFF
--- a/events/activities/pilgrimage_activity/pilgrimage_events.txt
+++ b/events/activities/pilgrimage_activity/pilgrimage_events.txt
@@ -2810,7 +2810,6 @@ scripted_trigger pilgrimage_2004_potential_friend_trigger = {
 	faith = root.faith
 	NOT = { is_courtier_of = root }
 	location = root.location
-	is_ruler = no #Unop To fix [ Trying to put character in court despite being a ruler ] (because of this char being taken in piligrimage travel)
 }
 
 #Run into a potential friend
@@ -2857,9 +2856,18 @@ pilgrimage.2004 = {
 			target = scope:potential_friend
 			modifier = divine_blessing_friendship_opinion
 		}
-		add_to_entourage_court_and_activity_effect = { 
-			CHAR_TO_ADD = scope:potential_friend 
-			NEW_COURT_OWNER = root 
+		if = { #Unop Don't poach a spouse, heir, or employed courtier of another ruler
+			limit = {
+				scope:potential_friend = { is_ruler = no }
+				can_recruit_character_to_court_trigger = {
+					RECRUITER = root
+					RECRUITEE = scope:potential_friend
+				}
+			}
+			add_to_entourage_court_and_activity_effect = {
+				CHAR_TO_ADD = scope:potential_friend
+				NEW_COURT_OWNER = root
+			}
 		}
 		hidden_effect = {
 			add_opinion = {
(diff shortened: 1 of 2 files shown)


pharaox:
"Invite to Court" without the fix:


"Invite to Court" with the fix:


pilgrimage.6001 with the fix (the friend is not joining your court):

events/activities/pilgrimage_activity/pilgrimage_events.txt
ProZeratul, line 7669: Seeing this event and I'm not liking that you can have your friend meeting you while you are traveling far far away. But on the other hand, preventing this would make this event likely to fire way less and the option would have no sense (since they would already be there)

ProZeratul: Could this be changed to use **drinks_alcohol_trigger**

ProZeratul (approved):

events/activities/pilgrimage_activity/pilgrimage_events.txt
pharaox: Yes, had similar thoughts, but decided against touching this. 2004 already requires same location, while 6001 doesn't, and adding it would mean it practically never fires. We could try teleporting them back and forth, but that gets complicated and won't work with rulers anyway.

pharaox: Good catch, changed it here and also in the event.

PR #473, opened 2026-05-16
Fix greek fire firing repeatedly and targeting the open Black Sea

### LINKED ISSUE #459: Greek fire fires on every sail-target change instead of once per month, and can target nonsense positions
**Reported on Steam by `White Lich` on `2026-05-05 20:32`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> Greek fire will fire every time the army sailing changes it's target instead of once per month per army and can target armies in nonsense positions like next to Crimea in the Black Sea.

pharaox:
Fixes #459

`ep3_emperor_yearly.3010` fires greek fire from `on_army_enter_province` (every province entry). Two vanilla bugs:

- **Fires every province entry, not once per 30 days.** The `used_fire_dromons` cooldown is checked in `.3010` on the invader (root), but set in `.3011` on the Byzantine defender (`scope:marmara_defender`) — so the check never matched and never blocked re-firing. Greek fire is one shared Byzantine fleet ("I cannot risk the fleet"; tooltip "Your Greek Fire ships will be unavailable for 30 days"), so the cooldown is per-defender. Now checked on `title:c_byzantion.holder`; the vanilla `.3011` set is left untouched. Declining (option B) intentionally sets no flag — the fleet wasn't spent.
- **Targets nonsense positions (Crimea / open Black Sea).** The `#strait of Marmara` location list included `province:950` and `province:8667` (`sea_black`, reaching Crimea). Removed; the actual strait approach (`947 sea_marmara`, `948 sea_gulf_varna`, `8665 sea_dardanelles`, `8668 sea_bosporus`) is kept.

### DIFF
--- a/events/dlc/ep3/ep3_emperor_yearly_3.txt
+++ b/events/dlc/ep3/ep3_emperor_yearly_3.txt
@@ -1,0 +1,1947 @@
+﻿namespace = ep3_emperor_yearly
+
+############################
+# EP3 Admin Emperor Events #
+# by Jason Cantalini   	   #
+# 3000 - 3500              #
+############################
+
+# Byzantine-only: your Varangians loot the palace after succession
+
+ep3_emperor_yearly.3000 = {
+	type = character_event
+	title = ep3_emperor_yearly.3000.t
+	desc = {
+		desc = ep3_emperor_yearly.3000.desc
+		first_valid = {
+			triggered_desc = {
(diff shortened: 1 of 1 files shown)


pharaox:
Verified in-game that there is a 30 day cooldown between events.

ProZeratul (approved):

PR #474, opened 2026-05-16
Fix AI never taking Request Royal Honsei due to influence cost

### LINKED ISSUE #462: Yamato clan AI not taking "Request Royal Honsei" decision
**Reported on Steam by `Darth Revan` on `2026-05-05 20:18`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> I don't know if you take bug report requests, but I've been running into an issue where members of the Yamato clan aren't taking the "Request Royal Honsei" decision, leading to the dynasty absolutely dominating Japan

pharaox:
Fixes #462

The CK3 AI never budgets a `cost`-block influence cost, so a decision with an influence cost and no `ai_goal` is silently never AI-executed even when every `is_shown`/`is_valid`/`ai_potential`/`ai_will_do` gate passes. 

Fix keeps influence a real constraint for the AI rather than making the decision free:
- `cost`: player cost unchanged.
- `is_shown`: AI only takes it when it can afford the influence.
- `effect`: AI actually pays the influence.

This engine quirk likely affects other vanilla influence-cost decisions, tracked in #475.

### DIFF
--- a/common/decisions/dlc_decisions/tgp/tgp_japan_decisions.txt
+++ b/common/decisions/dlc_decisions/tgp/tgp_japan_decisions.txt
@@ -838,8 +838,13 @@ tgp_japan_imperial_branch_decision = {
 		prestige = {
 			value = minor_prestige_gain
 		}
+		#Unop AI never budgets cost-block influence; charge it via is_shown/effect below
 		influence = {
-			value = minor_influence_gain
+			value = 0
+			if = {
+				limit = { is_ai = no }
+				add = minor_influence_gain
+			}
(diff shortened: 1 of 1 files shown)


pharaox:
Tested in-game that the proposed fix actually fixes the issue. 20 years into the game from 867 the decision was taken by AI 5 times, and the Yamato clan members and rulers stayed stable at ~60 / 3-4 during the entire period.

Also verified that the player is not affected by the fix.

ProZeratul (approved):

PR #484, opened 2026-05-23
Fix Random Harm game rule not affecting yearly trigger chance

### LINKED ISSUE #476: The "Random Harm" game rule is severely bugged
# Summary

I have grounds to believe the `Random Harm` game rule is currently bugged and does not work as intended. As in, there is no difference in yearly probability between `Illusion of Safety` (the lowest enabled) and `Tragically Spiteful` (the highest enabled), while `Safe` (fully disabled) may still trigger harm events as long as the character satisfies an optional condition of at least one.

# Explanation/Reasoning

The random harm events are currently being initiated as follows: `random_yearly_everyone_pulse` → `yearly_health_pulse` → `random_events` → `15 = harm.9501`

Because the `harm.9501` event does not have a `weight_multiplier` section, it always has a static/fixed chance to trigger, which is `15`. This event is then used to select an actual random harm event via the following line in its `immediate` section:

`trigger_event = { on_action = harm_events_pulse }`

The `harm_events_pulse` on_action has a `random_events` section with a bunch of weighted harm events. For example, `10 = harm.0501`. Each of these events has a `weight_multiplier` section, which in turn contains a modifier along the lines `modifier = { add = harm_game_rule_likelihood_value }`. The outcome of the value is based on the `Random Harm` game rule.

See what I'm getting at? The game rule DOES NOT at all affect the probability of triggering a harm event each year (what it should be doing according to description), but instead affects the probability of selecting a certain event itself once a random harm event successfully triggers (e.g., whether you will fall from a horse, slip in a bath, or choke on food). In other words, the yearly chance to trigger a harm event is unaffected. What is more, certain events may still trigger when the game rule is set to `Safe` if any other positive modifiers are present in their `weight_multiplier`!

# Primary fix suggestion

I would like to propose the following fix:

Add a `weight_multiplier` section to the `harm.9501` event, with the following contents:

```
	weight_multiplier = {
		base = 0
		modifier = { add = harm_game_rule_likelihood_value }
	}
```

As a result, the game rule will affect the chance of triggering a harm event each year.

The default outcome of the `harm_game_rule_likelihood_value` script value is 1 (when the game rule is set to `Dangerous`), so my fix preserves the intended base probability, while properly increasing or decreasing it depending on the game rule's setting.

# Optional second fix suggestion

As I explained, every harm event contains a modifier with `harm_game_rule_likelihood_value` in the `weight_multiplier` section. The primary fix makes these modifiers relatively harmless, preventing cases like events triggering when the game rule is set to `Safe`. However, the `harm_game_rule_likelihood_value` can be higher or lower than 1 depending on the `Random Harm` game rule setting. This leads to certain events becoming less or more likely to appear (relative to other harm events) depending on the setting, which does not seem intended.

To address this, I would like to propose replacing all `base = 0` with `base = 1` in the `weight_multiplier` sections inside the `game\events\harm_events.txt` file, while commenting out all the modifiers that contain `harm_game_rule_likelihood_value` (in events that use this scripted value for weighting in vanilla).

For example, this:
```
weight_multiplier = {
	base = 0
	# Add the weightings.
	modifier = { add = harm_game_rule_likelihood_value }
	# Weight up if you're _already_ inbred.
	modifier = {
		add = 0.5
		has_trait = inbred
	}
}
```
should turn into this:
```
weight_multiplier = {
	base = 1
	# Add the weightings.
	#modifier = { add = harm_game_rule_likelihood_value }
	# Weight up if you're _already_ inbred.
	modifier = {
		add = 0.5
		has_trait = inbred
	}
}
```

Once again, the default outcome of `harm_game_rule_likelihood_value` is 1 (when the game rule is set to `Dangerous`), so this fix preserves the harm events' intended relative probabilities across all game rule settings.

pharaox:
Fixes #476

`harm.9501` is the event that decides whether any random harm event fires in a given year (called at weight 15 from `yearly_health_pulse`). It had no `weight_multiplier`, so its yearly probability was independent of the `Random Harm` game rule — the rule only ever affected *which* harm event was selected once `harm.9501` had already fired, not the overall yearly chance.

Add a `weight_multiplier` to `harm.9501` that gates on `harm_game_rule_likelihood_value`, the same script value used by the individual harm events. This makes the yearly chance respect the game rule:

- `Illusion of Safety` → 0.25× the default chance
- `Dangerous` / `Spiteful` → default
- `Tragic` / `Tragically Spiteful` → 1.5× the default chance

### DIFF
--- a/events/harm_events.txt
+++ b/events/harm_events.txt
@@ -9681,6 +9681,12 @@ harm.9501 = {
 		is_incapable = no
 	}
 
+	#Unop Make the yearly chance to trigger a harm event respect the Random Harm game rule
+	weight_multiplier = {
+		base = 0
+		modifier = { add = harm_game_rule_likelihood_value }
+	}
+
 	immediate = {
 		trigger_event = { on_action = harm_events_pulse }
 	}


Nider001:
Just a heads up, I have added additional clarification in the issue post to explain the purpose of my second fix suggestion, which is currently not being implemented (with the reason why it should be implemented alongside the primary fix).

ProZeratul:
@Nider001 I'm merging this as a first step, we'll leave the issue open as I agree there is more work to do on this, just got to find the best way to do them (as there are other harm events in the health pulse)

PR #487, opened 2026-05-23
Document mandala court positions crash compatibility note

pharaox:
Add a short note to the Compatibility section of `README.md` and `description.bbc` flagging that if another mod overrides `common/court_positions/types/00_mandala_court_positions.txt` without the Unop fix of the `khlon_glan_court_position` aptitude, the game will crash, and advising users to ask the other mod's author to incorporate the fix.


ProZeratul:
Good idea, I'll also inform people that I know integrate unop into their mod

PR #488, opened 2026-05-26
Dev/fix raid estate insight

ProZeratul:
Raiding an estate should work for every type of holding, but it wasn't update for asian estates (except for the tea building)


ProZeratul:
The trigger part is done, now the effects have to be implemented too

PR #506, opened 2026-05-30
Keep relative harm event weights independent of Random Harm game rule

### LINKED ISSUE #476: The "Random Harm" game rule is severely bugged

pharaox:
Fixes #476

Secondary fix for #476 (the primary yearly-trigger-chance fix landed in #484).

For the 54 `harm_events_pulse` events, flip `base = 0`to `base = 1` and comment out the `harm_game_rule_likelihood_value` modifier. Since that value is 1 at the default Dangerous setting, the baseline weight is unchanged, but the relative ratios between events are now constant across all game-rule settings.

### DIFF
--- a/events/harm_events.txt
+++ b/events/harm_events.txt
@@ -913,10 +913,10 @@ harm.0501 = {
 	}
 
 	weight_multiplier = {
-		base = 0
+		base = 1 #Unop Keep relative harm event weights independent of the Random Harm game rule
 
 		# Add the weightings.
-		modifier = { add = harm_game_rule_likelihood_value }
+		#modifier = { add = harm_game_rule_likelihood_value }
 		# Weight up if you're _already_ inbred.
 		modifier = {
 			add = 0.5
@@ -1108,10 +1108,10 @@ harm.0502 = {
 	}
 
 	weight_multiplier = {
-		base = 0
+		base = 1 #Unop Keep relative harm event weights independent of the Random Harm game rule
 
 		# Add the weightings.
-		modifier = { add = harm_game_rule_likelihood_value }
+		#modifier = { add = harm_game_rule_likelihood_value }
 		# Weight up if you're _already_ inbred.
 		modifier = {
 			add = 0.5
@@ -1192,10 +1192,10 @@ harm.0511 = {
 	}
 
 	weight_multiplier = {
-		base = 0
+		base = 1 #Unop Keep relative harm event weights independent of the Random Harm game rule
 
 		# Add the weightings.
-		modifier = { add = harm_game_rule_likelihood_value }
+		#modifier = { add = harm_game_rule_likelihood_value }
 	}
 
 	immediate = {
@@ -1347,10 +1347,10 @@ harm.0512 = {
 	}
 
 	weight_multiplier = {
-		base = 0
+		base = 1 #Unop Keep relative harm event weights independent of the Random Harm game rule
 
 		# Add the weightings.
-		modifier = { add = harm_game_rule_likelihood_value }
+		#modifier = { add = harm_game_rule_likelihood_value }
 	}
 
 	immediate = {
@@ -1461,10 +1461,10 @@ harm.0521 = {
 	}
 
 	weight_multiplier = {
-		base = 0
+		base = 1 #Unop Keep relative harm event weights independent of the Random Harm game rule
 
 		# Add the weightings.
-		modifier = { add = harm_game_rule_likelihood_value }
+		#modifier = { add = harm_game_rule_likelihood_value }
 	}
 
 	immediate = {
@@ -1633,10 +1633,10 @@ harm.0522 = {
 	}
 
 	weight_multiplier = {
-		base = 0
+		base = 1 #Unop Keep relative harm event weights independent of the Random Harm game rule
 
 		# Add the weightings.
-		modifier = { add = harm_game_rule_likelihood_value }
+		#modifier = { add = harm_game_rule_likelihood_value }
 	}
 
 	immediate = {
@@ -1709,10 +1709,10 @@ harm.0531 = {
 	}
 
 	weight_multiplier = {
-		base = 0
+		base = 1 #Unop Keep relative harm event weights independent of the Random Harm game rule
 
 		# Add the weightings.
-		modifier = { add = harm_game_rule_likelihood_value }
+		#modifier = { add = harm_game_rule_likelihood_value }
 		# Plus some other factors apply.
 		modifier = {
 			add = 0.25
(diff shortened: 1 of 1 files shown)


pharaox:
I don't plan to verify this change in-game - there are too many events affected. The fix is straightforward, I think it should be ok to merge. Also relying on @Nider001 having verified it.

ProZeratul:
I aproved this MR, but I noticed that `commander_death_events_pulse `is also involving some harm events but is impacted by the same issues that previously fixed, I think I'll replicate what is done in `harm.9501` to mimic the call of all those commander harm events (also noticed they are not protecting players/AI from bad luck as they are not setting `harm_0000_not_on_dynastic_cooldown_trigger`)

ProZeratul (approved):

Nider001:
@ProZeratul the commander harm events do have a separate cooldown in the `on_army_monthly` on-action, from which they are triggered. And then these events themselves each have a 30 year cooldown in their definition. In other words, they are managed entirely separately from all other harm events. Interestingly, `harm.2051` and `harm.2061` are designed to never proc, because they have "always = no" in their trigger (I guess developers didn't have time to write the triggers, so they left placeholders instead - maybe look into this too?)

To clarify, are you going to create a new event to act as an intermediary for the commander harm events? E.g., like this (example draft, untested):
```
commander_death_events_pulse = {
	trigger = { is_commanding_army = yes }
	random_events = {
		1000 = 0
		120 = new_intermediary_event.1	# 20 * 8 - 20 * 2 (for two events that never trigger)
	}
}

new_intermediary_event.1 = {
	hidden = yes
	trigger = {
		harm_game_rule_enablement_trigger = yes	# maybe move to commander_death_events_pulse? (ruler/incapable checks like in harm.9501 are not needed, I think)
	}

	#Unop Make the yearly chance to trigger a harm event respect the Random Harm game rule
	weight_multiplier = {
		base = 0
		modifier = { add = harm_game_rule_likelihood_value }
	}

	immediate = {
		trigger_event = { on_action = unop_new_commander_death_events_pulse }
	}
}

unop_new_commander_death_events_pulse = {
	random_events = {
		20 = harm.2001	# Camp Fever - Terrible siege conditions cause you strife.
		20 = harm.2011	# Marching Malaise - Terrible marching conditions are problematic.
		20 = harm.2021	# So... Cold... - Frostbite claims you.
		20 = harm.2031	# So... Hot... - Heat-stroke claims you.
		20 = harm.2041	# For the Realm - Aggrieved soldiers at war too long assassinate you.
		20 = harm.2051	# Starvation - Complications due to starvation from lack of supply.
		20 = harm.2061	# Harried Each Step - Retreating from battle and caught by enemy soldiers.
		20 = harm.2071	# A Disastrous Victory - Pursuing too far ahead in battle and caught by enemy soldiers.
	}
}
```

ProZeratul:
@Nider001 indeed this was my intent. It will require some refinement to be sure everything is **harm**onized.

On `harm.2051` and `harm.2061` they in fact can now be implemented, it seems to be leftovers, from what I can tell they just need a check for supply state being starvation in `harm.2051` and something like
```
			involved_combat_side = {
				OR = {
					is_combat_side_retreating = yes
					enemy_side = { is_combat_side_pursuing = yes }
				}
			}
```
for harm.2061 (looks like this event is the reverse of harm.2071)

PR #507, opened 2026-05-30
Fix Swift Execution perk applying oversized murder scheme phase reduction

### LINKED ISSUE #504: "Swift Execution" perk enables LAAMPs to murder characters in no time at all
Choosing "Swift Execution" perk as a LAAMP turns you into an unstoppable and extremely quick murderer. It reduces the murder scheme phase duration by **100 days**. This means that you can murder most characters in just 5 * 10 = 50 days, even with low-ish intrigue.

The perk has:

```
	government_character_modifier = {
		flag = government_is_landless_adventurer
		murder_scheme_phase_duration_add = monumental_scheme_phase_duration_bonus_value
	}
```

This is the same code as for other perks that reduce befriend, seduce, elope, and courting scheme phase durations. But these "friendly" schemes have just a single very long phase (e.g. 1 year), so the reduction is by ~30%. The murder scheme on the other hand, as other hostile schemes, has multiple shorter phases, so the reduction approaches 100% and reduces the phase to the minimum duration of 10 days.

As this makes no sense at all, it should be regarded as a bug to be fixed. The `medium_scheme_phase_duration_bonus_value` for regular rulers should be similarly regarded as excessive, as again it's the same reduction as for the friendly schemes.

pharaox:
Fixes #504

The `swift_execution_perk` used `monumental` (-100 days) for LAAMPs and `medium` (-20 days) for regular rulers — the same values used by friendly scheme perks (befriend, seduce, courting, elope) against their single 365-day phase.

Fix: replace with values that give a proportionally equivalent reduction on murder's 120-day base:
- Regular ruler: `medium` (-20) → `minor` (-10, ~8%)
- LAAMP: `monumental` (-100) → `major` (-40, ~33%)

### DIFF
--- a/common/lifestyle_perks/00_intrigue_1_skulduggery_tree_perks.txt
+++ b/common/lifestyle_perks/00_intrigue_1_skulduggery_tree_perks.txt
@@ -1,0 +1,263 @@
+﻿truth_is_relative_perk = {
+	lifestyle = intrigue_lifestyle
+	tree = skulduggery
+	position = { 2 0 }
+	icon = node_intrigue
+
+	auto_selection_weight = {
+		value = 11
+		if = {
+			limit = {
+				has_education_intrigue_trigger = yes
+			}
+			add = 1989
+		}
+		if = {
+			limit = {
+				has_focus = intrigue_skulduggery_focus
+			}
+			multiply = 5
+		}
+		if = {
+			limit = {
+				can_start_new_lifestyle_tree_trigger = no
+			}
+			multiply = 0
+		}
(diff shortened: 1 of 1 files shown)


ProZeratul:
It's been like that since RtP (1.13 patch) indeed, so it match the refactoring of schemes and as you said, probably that there was some confusion with how they were now working.

Previously the bonus was `murder_scheme_power_mult = 0.3` but it doesn't help us there.

I agree this reduction of 100 days on the 120 is way too much.

The only thing here is to be prepared to be pitchforked by users being mad that they lost their minigun

ProZeratul (approved):

PR #513, opened 2026-06-05
Fix admin noble families keeping obligation theme after losing governorship

### LINKED ISSUE #498: Administrative noble families keep obligation type after losing governorship
**Reported on Steam by `White Lich` on `2026-05-27 19:11`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> In Administrative governments, Noble Families get the obligation type of their governorship. When they lose their governorship, they will keep their obligation type despite not having any land. If that obligation type has a military budget, they will take the military budget and have it vanish because they don't have a treasury. Seems to work fine for Celestial and Meritocratic though.

Split out from #494.

pharaox:
Fixes #498

In Administrative governments, Noble Families inherit the obligation theme of their governorship. When they lose their governorship, they keep that theme despite no longer holding land. For `admin_theme_military` (and `admin_theme_civilian`), which lack any `is_valid` guard, the theme persists indefinitely; `admin_theme_military`'s `monthly_treasury_from_military_budget_base` modifier then keeps firing into nothing.

Administrative noble-family titles are `tier_duchy`, so the group-level `is_shown = { scope:subject.primary_title.tier >= tier_duchy }` never becomes false (unlike Celestial/Meritocratic, where the NF title is county-tier or the group checks `is_landless_ruler`). The frontier/naval themes already self-invalidate via their border/coastal `is_valid` checks.

This adds an `is_valid` guard to `admin_theme_military` and `admin_theme_civilian` requiring the subject to hold a real (non-noble-family) duchy-tier governorship — reusing the exact idiom already used by `administrative_salary_rank_duchy` in the same file. When the family loses its governorship, the theme falls back to the default `admin_theme_balanced`.

### DIFF
--- a/common/subject_contracts/contracts/administrative.txt
+++ b/common/subject_contracts/contracts/administrative.txt
@@ -88,6 +88,15 @@ administrative_themes = {
 			score = 0
 			color = { 0.2 0.2 0.7 1.0 }
 
+			#Unop Reset to the default theme when the noble family loses its governorship (only holds its noble family title)
+			is_valid = {
+				scope:subject = {
+					any_held_title = {
+						is_noble_family_title = no
+					}
+				}
+			}
+
 			tax_factor = 0.5
 
 			subject_modifier = {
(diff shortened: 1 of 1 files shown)


pharaox:
Wrong fix, will be done again.

PR #517, opened 2026-06-07
Fix restore_backwater_counties_decision: AI never takes it, governor-tier gate

### LINKED ISSUE #501: Restore Backwater Counties decision: AI never takes it; emperor tier vassals incorrectly excluded
**Reported on Steam by `White Lich` on `2026-05-30 03:33`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> The AI will never take the Restore Backwater Counties decision because it has no AI weights, also emperor tier vassals aren't allowed to take it even though they should.

Split out from #495.

pharaox:
Fixes #501

**fixer:** Two vanilla bugs in `restore_backwater_counties_decision` (`common/decisions/dlc_decisions/ep3_decisions.txt`):

1. **AI never takes the decision** — it had no `ai_potential`/`ai_will_do`, the only `decision_group_type = roman` decision lacking them. Added both (modeled on siblings), and raised `ai_check_interval_by_tier` `empire`/`hegemony` from `0` to `120` so the AI checks it for empire-tier vassal governors.
2. **Governors wrongly excluded by a tier proxy** — `is_shown` gated `primary_title` to duchy/kingdom tier. That tier gate is a *proxy* for "is a vassal governor (has a liege)"; the decision's `effect` notifies `liege = { ... }` and its loc reads "Restore [PrimaryTitle]", so it structurally needs a liege. Replaced the proxy with the real condition, `is_governor = yes` (which includes `top_liege != this`). This admits empire-tier vassal governors (e.g. under a hegemony) while still excluding an independent empire-tier emperor — avoiding the broken loc / `liege = {}` that a blanket `tier = tier_empire` would cause.

**Needs in-game verification:** AI governors actually take the decision at sane cadence; an empire-tier vassal governor (hegemony setup) now sees it; an independent empire-tier emperor still does **not**.

### DIFF
--- a/common/decisions/dlc_decisions/ep3_decisions.txt
+++ b/common/decisions/dlc_decisions/ep3_decisions.txt
@@ -115,7 +115,7 @@ restore_backwater_counties_decision = {
 		county = 0
 		duchy = 120
 		kingdom = 120
-		empire = 0
+		empire = 120 #Unop Let the AI check this decision for empire-tier vassal governors
 		hegemony = 0
 	}
 
@@ -127,12 +127,13 @@ restore_backwater_counties_decision = {
 
 	is_shown = {
 		has_ep3_dlc_trigger = yes
-		primary_title ?= {
-			OR = {
-				tier = tier_duchy
-				tier = tier_kingdom
-			}
-		}
+		is_governor = yes #Unop Gate on being a governor (vassal with a liege) instead of duchy/kingdom tier, so empire-tier vassal governors qualify while independent emperors stay excluded
+		#primary_title ?= {
+		#	OR = {
+		#		tier = tier_duchy
+		#		tier = tier_kingdom
+		#	}
+		#}
 		any_sub_realm_county = {
 			has_county_modifier = backwater_county_modifier
 		}
@@ -272,6 +273,26 @@ restore_backwater_counties_decision = {
 				# Trigger the event for the emperor to see if all the counties have been cleared
 				trigger_event = ep3_decisions_event.3001
 			}
+		}
+	}
+
+	ai_will_do = { #Unop Give the AI weights so governors actually take this decision
+		base = 20
+		modifier = {
+			has_trait = ambitious
+			add = 20
+		}
+		modifier = {
+			has_trait = diligent
+			add = 10
+		}
+		modifier = {
+			has_trait = content
+			add = -10
+		}
+		modifier = {
+			has_trait = lazy
+			add = -10
 		}
 	}
 }


pharaox:
**reviewer:** Looks good — both bugs fixed at the root cause; tiger clean (0/0).

- `is_governor = yes` correctly admits empire-tier vassal governors while excluding independent emperors (`top_liege != this`); old tier proxy commented out.
- `ai_potential`/`ai_will_do` added (modeled on siblings) and `empire`/`hegemony` interval raised 0→120, so the AI both checks and weights the decision. `ai_potential`'s broader `government_has_flag` gate is harmless since `is_shown` still carries `is_governor`.
- No findings. In-game verification still recommended: AI cadence, empire-tier vassal governor sees it, independent emperor does not.

common/decisions/dlc_decisions/ep3_decisions.txt
```diff
@@ -274,6 +275,30 @@ restore_backwater_counties_decision = {
 			}
 		}
 	}
+
+	ai_potential = { #Unop Give the AI weights so governors actually take this decision
+		government_has_flag = government_is_administrative
+	}
```
pharaox: Is this needed?

common/decisions/dlc_decisions/ep3_decisions.txt
```diff
@@ -115,8 +115,8 @@ restore_backwater_counties_decision = {
 		county = 0
 		duchy = 120
 		kingdom = 120
-		empire = 0
-		hegemony = 0
+		empire = 120 #Unop Let the AI check this decision for empire-tier vassal governors
+		hegemony = 120 #Unop Let the AI check this decision for hegemony-tier vassal governors
```
pharaox:
```suggestion
		hegemony = 0
```
There is no such thing as "hegemony-tier vassal governors". A hegemon is the highest possible tier, never a vassal.

common/decisions/dlc_decisions/ep3_decisions.txt
```diff
@@ -127,12 +127,13 @@ restore_backwater_counties_decision = {
...
-				tier = tier_kingdom
-			}
-		}
+		is_governor = yes #Unop Gate on being a governor (vassal with a liege) instead of duchy/kingdom tier, so empire-tier vassal governors qualify while independent emperors stay excluded
+		#primary_title ?= {
+		#	OR = {
+		#		tier = tier_duchy
+		#		tier = tier_kingdom
+		#	}
+		#}
```
pharaox, line 136: Gating on `is_governor` looks correct, however removing the primary title tier gate could potentially open up the decision for counts as well - double check.

pharaox:
`@fixer` Added inline comments, check.

common/decisions/dlc_decisions/ep3_decisions.txt
pharaox: **fixer:** No — removed it. `ai_potential` is only a cheap pre-filter for *whether the AI looks*; the required AI behavior comes from `ai_will_do` + `ai_check_interval_by_tier`. My gate (`government_has_flag = government_is_administrative`) was strictly broader than `is_shown`'s `is_governor = yes` (which already implies `government_allows = administrative`), so it filtered nothing. Dropped.

common/decisions/dlc_decisions/ep3_decisions.txt
pharaox: **fixer:** Agreed, reverted `hegemony` to `0`. A hegemon is the top tier and never a vassal, so it can never satisfy `is_governor` (`top_liege != this`) — checking it there is pointless. Also corrected the now-misleading `empire` comment.

common/decisions/dlc_decisions/ep3_decisions.txt
pharaox: **fixer:** Double-checked: counts stay excluded. `is_governor` floors the tier via `OR = { highest_held_title_tier >= main_administrative_tier; highest_held_title_tier >= min_title_maa_tier }`, and `main_administrative_tier` resolves to `duchy` for administrative governments. The separate `is_governor_or_admin_count` trigger exists precisely to *include* counts (`>= tier_county`) — we deliberately use the count-excluding `is_governor` here.

pharaox:
**fixer:** Addressed all three inline comments (amended the existing commit; tiger clean 0/0):

- **`ai_potential` removed** — it was a no-op pre-filter (broader than `is_shown`'s `is_governor`), and AI behavior is fully driven by `ai_will_do` + `ai_check_interval_by_tier`.
- **`hegemony` reverted to `0`** — a hegemon is the top tier and can never be a vassal governor (`is_governor` requires `top_liege != this`). Fixed the misleading `empire` comment too.
- **Counts confirmed excluded** — `is_governor` floors at duchy (`main_administrative_tier = duchy`); the count-including variant is the separate `is_governor_or_admin_count`. No change needed.

In-game verification still recommended: AI governors take it on cadence; an empire-tier vassal governor sees it; an independent administrative emperor does not.

pharaox:
Verified in-game that all 3 governor tiers see the decision, while emperor and hegemon top-lieges don't. Haven't verified that AI actually uses it but the fix looks straightforward enough.

pharaox:
**reviewer:** Looks good — re-fix correctly addresses all three inline comments; tiger clean (0/0).

- `ai_potential` removal is sound (it was strictly broader than `is_shown`'s `is_governor`, so a no-op pre-filter); AI is driven by `ai_will_do` + `ai_check_interval_by_tier`.
- `hegemony = 0` revert is correct (a hegemon can never satisfy `top_liege != this`); `empire = 120, hegemony = 0` matches sibling patterns in the same file.
- Counts stay excluded — `is_governor` floors at duchy via `main_administrative_tier`; the count-including variant `is_governor_or_admin_count` is deliberately not used.
- Maintainer's in-game verification (all 3 governor tiers see it; emperor/hegemon top-lieges don't) confirms the core gate behavior. No findings.

ProZeratul:
Since it was before Hegemon introduction I think this is good to go, without hegemon being a thing, this would not have sense (in the way the top emperor would have other things to do)

PR #529, opened 2026-06-14
Fix peasant_leader trait swap when remaining landless after winning separatist uprising

### LINKED ISSUE #528: Peasant Leader trait wrongly swapped for Populist Leader when aiding peasants without becoming landed
**Reported on Steam by `Ryūichi` on `2026-06-13 06:25`** ([link](https://steamcommunity.com/workshop/filedetails/discussion/2871648329/4697908845434539387/)).

> Apologies in advance if this was already pointed out.
>
> There's what I believe is a bug when it comes to gaining and progressing the Peasant Leader trait when playing as an Adventurer.
>
> If you take a 'Fight Corruption' contract and side with the peasants instead of the lord during the various events, you have at the end the option (among others) to side with the peasantry and trigger / join a peasant rebellion.
>
> When given that option, you can then choose whether to become landed there if you win the war (top option out of four, I believe) or just aid the peasants and staying an adventurer (2nd option).
>
> The issue is that if you selected the 2nd option (simply aiding the peasants and not becoming landed), the game correctly gives you the Peasant Leader trait (this is the one that can be leveled) at war start, but then incorrectly removes it at the end of the war and replaces it with the Populist Leader trait (the one without levels), which you're only supposed to get when you opt to get landed by picking the first option in the previous event.
>
> I'll list the repro steps below if it's not clear
>
> 1. Play as an adventurer
> 2. Take a 'Fight Corruption' contract (might need some in game time as it usually requires low county control; switching camp purpose to Swords-for-Hire might speed things up)
> 3. During the events that follow (can be a single one too), always pick the bottom option of the two, the one to side with the peasants
> 4. When the final event triggers, choose the 2nd option out of four (the one to aid the peasants but not become landed) and observe as the Peasant Leader leveled trait is given to you
> 5. Win the war, don't enforce demands
> 6. Before enforcing demands, observe how the tooltip there tells you that you'll seize all territory etc (remember that you picked the option not to and stay unlanded) and that it'll give you the Populist Leader trait and remove the Peasant Leader one
>
> Thank you guys for all your work with Unop!

pharaox:
Fixes #528

**fixer:** In `ep3_laamp_peasant_war`'s `on_victory`, the swap of the levelable `peasant_leader` trait for `populist_leader` ran unconditionally on the attacker. When the player chose to aid the peasants but stay landless (`ep3_laamp_peasant_war_remain_laamp` set), this still stripped `peasant_leader` — their only way to level it via these contracts — even though the rest of `on_victory` already gates landed-vs-landless behavior on that variable.

The fix wraps the two trait lines in an `if` gated on `NOT = { has_variable = ep3_laamp_peasant_war_remain_laamp }`, so the swap only happens when the player actually becomes landed.

Left `set_variable = became_landed_through_separatist_uprising_var` ungated: it only suppresses the `ep3_laamps.0800` "you became landed" event, whose trigger requires `is_landed = yes`; on a remain-landless win the player isn't landed, so that event never fires regardless — gating it would be an unnecessary change.

### DIFF
--- a/common/casus_belli_types/07_ep3_wars.txt
+++ b/common/casus_belli_types/07_ep3_wars.txt
@@ -6052,11 +6052,17 @@ ep3_laamp_peasant_war = {
 				save_scope_as = old_title
 			}
 			show_pow_release_message_effect = yes
-			#set this to not display the event to keep playing as landless
-			set_variable = became_landed_through_separatist_uprising_var
-			#swap the peasant trait for the better one 
-			remove_trait = peasant_leader
-			add_trait = populist_leader
+			#Unop only set the variable and swap the trait when actually becoming landed
+			if = {
+				limit = {
+					NOT = { has_variable = ep3_laamp_peasant_war_remain_laamp }
+				}
+				#set this to not display the event to keep playing as landless
+				set_variable = became_landed_through_separatist_uprising_var
+				#swap the peasant trait for the better one
+				remove_trait = peasant_leader
+				add_trait = populist_leader
+			}
 			if = {
 				limit = {
 					has_variable = ep3_laamp_peasant_war_remain_laamp


common/casus_belli_types/07_ep3_wars.txt
```diff
@@ -6054,9 +6054,15 @@ ep3_laamp_peasant_war = {
 			show_pow_release_message_effect = yes
 			#set this to not display the event to keep playing as landless
 			set_variable = became_landed_through_separatist_uprising_var
```
pharaox: **reviewer:** Low / pre-existing — leaving `set_variable` ungated is correct for fixing this issue (it's a no-op on a landless win, since `ep3_laamps.0800` needs `is_landed = yes`). Noting only that this var is never cleared anywhere, so a remain-landless winner keeps it forever; if they later become landed by another route, `ep3_laamps.0800` would be suppressed. Vanilla has the same quirk, so out of scope here — flagging for the tester, not asking for a change.

pharaox (commented):
**reviewer:** Looks good — the swap is now gated on `NOT = { has_variable = ep3_laamp_peasant_war_remain_laamp }`, the inverse of the adjacent landless branch, so it fires only when the player actually becomes landed. Minimal, root-cause, reuses the existing variable; tiger re-run clean. Verified the ungated `became_landed_through_separatist_uprising_var` is a no-op here (its only consumer, `ep3_laamps.0800`, requires `is_landed = yes`).

- Low (pre-existing, not a regression): that variable is never cleared anywhere, so a remain-landless winner carries it permanently; if they later become landed via another path, the `ep3_laamps.0800` notification would be suppressed. Out of scope for this issue — vanilla behaves identically — but worth a tester note.

Please confirm in-game: a remain-landless win keeps the levelable `peasant_leader`; a become-landed win still swaps to `populist_leader`.

pharaox:
`@fixer` Address the reviewer comment above. It's real issue, albeit a different one from what was reported.

pharaox:
**fixer:** Addressed the reviewer's finding. `set_variable = became_landed_through_separatist_uprising_var` was set unconditionally on `on_victory`, so a remain-landless winner kept it forever — and since the var is never cleared anywhere, `ep3_laamps.0800` would later be wrongly suppressed if that adventurer became landed by another route.

Moved the `set_variable` into the same `NOT = { has_variable = ep3_laamp_peasant_war_remain_laamp }` guard that gates the trait swap, so both effects apply only when actually becoming landed — matching how the rest of `on_victory` routes landed/landless behavior. The other set site (`ep3_decisions_events.txt`) is on a genuine become-landed decision and is left as-is.

`make tiger` clean (0/0/0). Amended into the single commit on the branch.

common/casus_belli_types/07_ep3_wars.txt
pharaox: **fixer:** Fixed — moved the `set_variable` inside the become-landed guard alongside the trait swap, so a remain-landless winner no longer carries the stale `became_landed_through_separatist_uprising_var`.

common/casus_belli_types/07_ep3_wars.txt
ProZeratul, line 6208: This effect is run on the `peasant_leader ` scope checking for `ep3_laamp_peasant_war_remain_laamp` but if it is set, is destroying the LAAMP status of `scope:attacker` then try to return home

This should probably be a NOT before the variable check and destroy the peasant leader camp and not the attacker

ProZeratul (changes requested):

common/casus_belli_types/07_ep3_wars.txt
pharaox: This indeed makes no sense, but looking at how `scope:peasant_leader` is initialized, this seems to be dead code, so I wouldn't touch it. Adding a NOT will most likely cause issues - the non-LAAMP peasant leader will not have the variable, so the real LAAMP will get destroyed when they shouldn't.

`@fixer` Please analyze this and advise.

pharaox: **fixer:** Analyzed — agreed with you, this block is dead code and should be left as-is. A `NOT` would actively break both outcomes.

**Who `scope:peasant_leader` is, and who holds the variable**

`ep3_laamp_peasant_war_remain_laamp` is set in exactly one place — `ep3_contract_event.0095` option `a_2`, on the event `root` (the player/attacker who chose to win but stay landless). It is **only ever on `scope:attacker`**, never on any other character.

`scope:peasant_leader` in `on_victory` is assigned by the branch just above (lines 6066-6105):
- **Remain-laamp** (attacker *has* the var): `scope:peasant_leader` = a *different* peasant from `war_met_peasant_leaders_list` with `this != scope:attacker`, or a freshly `create_character`'d NPC. Never the attacker, and these NPCs don't carry the var.
- **Become-landed** (`else`, attacker does *not* have the var): `scope:peasant_leader = scope:attacker`, who by definition lacks the var in this branch.

So in **both** paths `scope:peasant_leader` never has `ep3_laamp_peasant_war_remain_laamp` → the `limit` at line 6203 is always false → the `destroy_laamp_effect` + `return_home` never run. Dead code, and it's byte-identical to vanilla (this PR didn't touch it).

**Why adding `NOT` is wrong**

A `NOT` makes the limit always *true*, firing in both paths:
- Remain-laamp: `scope:peasant_leader` is the NPC who just took the land, but the body destroys `scope:attacker`'s LAAMP and sends the NPC home — tearing down the realm that was just created and hitting the wrong character, exactly as you predicted.
- Become-landed: `scope:peasant_leader` = attacker, so it would strip the LAAMP and return-home the player who was supposed to become landed — undoing the whole victory.

The `ADVENTURER = scope:attacker` hardcoding only makes sense if `scope:peasant_leader` were always the attacker, which it isn't. The author likely intended the body to act on `scope:peasant_leader` itself, but since the limit never passes it was never exercised.

**Advice:** leave the block untouched — out of scope for #528, vanilla-identical, and any change here risks the breakage above. If we ever want to clean it up it should be a separate issue with in-game repro, not bundled into this fix.

ProZeratul (approved):

PR #541, opened 2026-06-20
Fix ep3_contract_event.0007 city buy assigning bought artifact to merchant

### LINKED ISSUE #404: ep3_contract_events assigns artifact to wrong owner

pharaox:
Fixes #404

**fixer:** In the treasure-hunt contract resolution event `ep3_contract_event.0007`, the city branch reveals a second artifact that the merchant offers to sell, and the "I'll take everything" option (`city_city`) has the player pay for it. But after `change_treasure_type` correctly transfers it to the player, the option re-assigned it back to the merchant with `set_owner = scope:local_1`, so the player paid and got nothing. Changed that to `set_owner = root` so the bought artifact ends up with the player.

### DIFF
--- a/events/dlc/ep3/ep3_contract_events.txt
+++ b/events/dlc/ep3/ep3_contract_events.txt
@@ -1957,7 +1957,7 @@ ep3_contract_event.0007 = {
 		}
 		change_treasure_type = yes
 		scope:newly_created_artifact = {
-			set_owner = scope:local_1
+			set_owner = root #Unop the player bought the artifact, so it should go to them, not back to the merchant
 		}
 
 		stress_impact = {


events/dlc/ep3/ep3_contract_events.txt
```diff
@@ -1957,7 +1957,7 @@ ep3_contract_event.0007 = {
 		}
 		change_treasure_type = yes
 		scope:newly_created_artifact = {
-			set_owner = scope:local_1
+			set_owner = root #Unop the player bought the artifact, so it should go to them, not back to the merchant
```
pharaox, line 1960: **reviewer:** Correct. At this point `scope:newly_created_artifact` is artifact B (the merchant's revealed one — the immediate's second `inspiration_adventure_create_artifact_effect` repointed the scope to it), which is the artifact named in `desc_city_city` and offered as part of "sell both". The player has already paid `local_1` above, so `root` is the right owner. Verified `make tiger` is clean.

pharaox (commented):
**reviewer:** Looks good — the `set_owner = scope:local_1` → `set_owner = root` swap correctly gives the bought artifact (B, the one the merchant reveals and names in the desc, offered as "sell both") to the player instead of back to the merchant. Verified the scope: `inspiration_adventure_create_artifact_effect` in the immediate repoints `scope:newly_created_artifact` to B via its inner `create_artifact { save_scope_as = newly_created_artifact }`, so this option operates on B; the "I'll take everything" fiction matches transferring it to root. Other `set_owner = scope:local_1` lines in the file are lost-contest / give-back-to-locals cases, not the same bug. `make tiger` clean.

- Low: the secondary reported symptom (extra "received artifact" notification / duplicate artifact-history entries from the double create in the immediate) is untouched — correctly out of scope for this minimal fix, but worth confirming in-game.

ProZeratul:
I'm merging this as it's fine by itself, but I'll leave my other PR open until I get a closure with my gut feeling

PR #544, opened 2026-06-20
Gate faltering_heart xp in stress_threshold.0001 by stress level

### LINKED ISSUE #537: Faltering Heart gains experience even at stress level 0
**Reported on Steam by `Nider001` on `2026-06-16 17:26`** (https://steamcommunity.com/workshop/filedetails/discussion/2871648329/4697908845434539387/).

> The description of Faltering Heart explicitly states that it progresses every mental break. However, I've seen this trait gain experience even at stress level 0, which is clearly a bug.
>
> From what I found, when the mental break cooldown expires (handled via "stress_threshold.0005"), it triggers the "stress_threshold.0001" and "stress_threshold.0002" events regardless of current stress. The "stress_threshold.0002" correctly checks if the character has any stress level before adding exp to Faltering Heart, but "stress_threshold.0001" does not (hence the unintended trait level ups). For the record, Faltering Heart does not gain exp anywhere else in the base game scripts.
>
> I also find it weird that "stress_threshold.0002" has different amounts of trait exp gain for level 1/2/3 mental breaks, while "stress_threshold.0001" does not. Might be an oversight by devs.
>
> Playtested solution to address everything described above: replace the Faltering Heart block in "stress_threshold.0001" with the respective block from "stress_threshold.0002".

pharaox:
Fixes #537

**fixer:** In `stress_threshold.0001` (the non-imprisoned periodic stress event), the Faltering Heart xp block had no stress gate — it always added a flat 25 track xp whenever the event fired, including at stress level 0. Since `stress_threshold.0005` fires `.0001` on cooldown regardless of stress, the trait progressed (and could reach the heart-attack death at xp ≥ 100) with zero stress, contradicting the trait's "progresses for every mental break" design.

Mirrored the imprisoned path (`stress_threshold.0002`), which already tiers the xp by stress level: 25 at `stress_level >= 3`, 15 at `>= 2`, 5 at `>= 1`, and nothing at level 0. The ungated death check is left unchanged. This also resolves the flat-vs-tiered inconsistency between the two paths.

The vanilla file is added unmodified in the first commit, the fix in the second. In-game verification needed: at stress 0 the trait should no longer gain xp.

### DIFF
--- a/events/stress_events/stress_threshold_events.txt
+++ b/events/stress_events/stress_threshold_events.txt
@@ -1,0 +1,17534 @@
+﻿namespace = stress_threshold
+
+###############
+# Hidden Stress Framework Events
+#	stress_threshold.0001 — Fired when character crosses a stress threshold. Checks if a Mental Break event can be fired and, if yes, does so.
+#	stress_threshold.0002 — As above, but for characters in prison (they get events from a different pool, for obvious reasons).
+#	stress_threshold.0005 — Fired 5 (or 8, with a perk) years from the last Mental Break a character received. Re-fires either 0001 or 0002, as appropriate.
+#	stress_threshold.0010 — Sends a toast to players when they gain a Stress Level.
+#	stress_threshold.0400 — Fired when a character reaches 400 Stress. Resets the normal 5-year cooldown for Mental Breaks and immediately fires ones.
+#
+# Level 1 Mental Breaks
+#	stress_threshold.1001 — Dark Thoughts: Suffering from overwhelming feelings of guilt.
+#	stress_threshold.1011 — Wanton Desires: Distracted by lavacious fantasies.
+#	stress_threshold.1501 —	Nothing Matters: Depression-like disinterest in everything.
+#	stress_threshold.1502 — Resentment: Increasingly frustrated and irritated with everyone.
+#	stress_threshold.1503 —	Emptiness Inside: Ennui
+#	stress_threshold.1504 —	Endless Toil: Suffering from burnout.
+#	stress_threshold.1601 — Spinning Thoughts [Generic]: Overwhelmed by the number of things to keep track of.
+#	stress_threshold.1602 —	Swirling Emotions [Generic]: Struggling to keep emotions in-check.
+#	stress_threshold.1701 — Prior Error: your personality has caused friction elsewhere.
+#	stress_threshold.1711 — Too Busy: everything just seems to pile up on you these days.
+#	stress_threshold.1721 — Impostor Syndrome: clearly you just aren't good enough for your rank!
+#	stress_threshold.1731 — No Time for Myself: your limited time to yourself is intruded on.
+#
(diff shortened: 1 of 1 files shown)


events/stress_events/stress_threshold_events.txt
```diff
@@ -10797,12 +10825,12 @@ stress_threshold.2321 = {
...
-				has_two_stress_threshold_options = no
-				is_alive = yes
-				has_trait = irritable	
-			}
+		if = { #Irritable (tends to lash out at random)
+			limit = {
+				has_two_stress_threshold_options = no
+				is_alive = yes
+				has_trait = irritable	
+			}
```
pharaox: Unrelated whitespace change, revert.

pharaox (commented):
**reviewer:** Looks good — the faltering_heart block in `stress_threshold.0001` is now an exact mirror of `stress_threshold.0002` (tiered 25/15/5 gated by `stress_level >= 3/2/1`, nothing at 0; death check left ungated), which fixes the stress-0 xp gain and the flat-vs-tiered inconsistency in one change. Commit 1 is byte-identical to vanilla; the lone tiger warning is the pre-existing `hunt_create_trophy_effect` one (present on `main`, unrelated), so no new warnings.

- Low: commit 2 also has an unrelated LF→CRLF change at `stress_threshold.2321` (~10825). It should be reverted so the commit only touches the faltering_heart block — agreeing with @pharaox's inline note "Unrelated whitespace change, revert." No functional impact.

events/stress_events/stress_threshold_events.txt
```diff
@@ -0,0 +1,17562 @@
...
+		}
+		else = {
+		}
+		if = {
+			limit = {
+				has_trait = faltering_heart
+				is_incapable = no
+			}
+			#Unop gate the faltering_heart xp by stress level (and tier it) like the imprisoned path does, so it no longer progresses at stress level 0
```
pharaox, line 198: **reviewer:** Correct — this is a line-for-line mirror of the imprisoned path `stress_threshold.0002`, and the ungated `xp >= 100 → death_heart_attack` check is preserved below the tiered chain. At stress 0 no branch fires, so the trait no longer progresses, matching the "progresses for every mental break" design.

pharaox:
**fixer:** Reverted the unrelated line-ending change at `stress_threshold.2321` (the irritable block's 6 LF lines that had been normalized to CRLF). The fix commit now diffs against vanilla as the faltering_heart block only — verified the file matches the unmodified vanilla everywhere else. Amended and force-pushed; `make tiger` clean.

events/stress_events/stress_threshold_events.txt
pharaox: **fixer:** Reverted — those 6 lines (the irritable `if`/`limit` block) are back to vanilla LF endings, so the commit now touches only the faltering_heart block.

PR #545, opened 2026-06-20
Change coronation law before equipping crown/regalia

### LINKED ISSUE #538: Coronation fails to auto-equip crown/regalia due to law-change ordering
**Reported on Steam by `Nider001` on `2026-06-16 18:10`** (https://steamcommunity.com/workshop/filedetails/discussion/2871648329/4697908845434539387/).

> There is also a bug in the "coronation_events_6.txt" file related to equipping the crown/regalia during coronation. Right now the relevant script looks like this (similar script is used a total of six times in the file):
> if = {
> 	limit = {
> 		scope:crowning_artifact ?= { is_equipped = no }
> 		can_equip_artifact = scope:crowning_artifact
> 	}
> 	scope:crowning_artifact = { equip_artifact_to_owner = yes }
> }
> coronation_change_law_effect = yes
> As you can see, "can_equip_artifact = scope:crowning_artifact" is checked before the law is changed via "coronation_change_law_effect = yes". This is incorrect, because characters can never equip crown/regalia while having the Uncrowned law (I have to manually equip the artifact every coronation).
>
> Playtested solution (tested in two different coronations - the crown got automatically equipped both times): alter all script occurrences in the file to attempt equipping the artifact AFTER the law is changed. Like this:
> coronation_change_law_effect = yes
> if = {
> 	limit = {
> 		scope:crowning_artifact ?= { is_equipped = no }
> 		can_equip_artifact = scope:crowning_artifact
> 	}
> 	scope:crowning_artifact = { equip_artifact_to_owner = yes }
> }

pharaox:
Fixes #538

**fixer:** During coronation the crown/regalia auto-equip ran `can_equip_artifact` *before* `coronation_change_law_effect = yes`. While the ruler still holds the `uncrowned` law, the crown/regalia templates' `can_equip` (`NOT = { has_realm_law = uncrowned }`) fails, so the equip `if` is skipped and the artifact is never equipped automatically.

Reordered so `coronation_change_law_effect = yes` runs first at all six sites in `coronation_events_6.txt` (the king/emperor/anointment effect blocks and their `show_as_tooltip` mirrors), so the law no longer blocks the equip and the tooltip matches the executed effect.

### DIFF
--- a/events/activities/coronation_activity/coronation_events_6.txt
+++ b/events/activities/coronation_activity/coronation_events_6.txt
@@ -5312,6 +5312,7 @@ coronation_events.6100 = {
 			limit = {
 				this = scope:host
 			}
+			coronation_change_law_effect = yes #Unop Change the law first, so the crown/regalia can be equipped
 			if = {
 				limit = {
 					scope:crowning_artifact ?= { is_equipped = no }
@@ -5319,11 +5320,11 @@ coronation_events.6100 = {
 				}
 				scope:crowning_artifact = { equip_artifact_to_owner = yes }
 			}
-			coronation_change_law_effect = yes
 		}
 		else = {
 			show_as_tooltip = {
 				scope:host = {
+					coronation_change_law_effect = yes #Unop Change the law first, so the crown/regalia can be equipped
 					if = {
 						limit = {
 							scope:crowning_artifact ?= { is_equipped = no }
@@ -5331,7 +5332,6 @@ coronation_events.6100 = {
 						}
 						scope:crowning_artifact = { equip_artifact_to_owner = yes }
 					}
-					coronation_change_law_effect = yes
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


events/activities/coronation_activity/coronation_events_6.txt
```diff
@@ -5312,26 +5312,26 @@ coronation_events.6100 = {
 			limit = {
 				this = scope:host
 			}
+			coronation_change_law_effect = yes #Unop Change the law first, so the crown/regalia can be equipped
```
pharaox, line 5315: **reviewer:** Correct. `coronation_change_law_effect` here runs in the host scope (gated by `this = scope:host`), replacing `uncrowned` with `crowned_king`/`crowned_emperor` before the equip check, so `can_equip_artifact = scope:crowning_artifact` now passes. Same reasoning applies to the tooltip mirror just below and the 6101/6102 sites.

pharaox (commented):
**reviewer:** Looks good — moving `coronation_change_law_effect = yes` before the equip `if` at all 6 sites is the correct, minimal fix. While `uncrowned` is still active the templates' `can_equip = NOT { has_realm_law = uncrowned }` fails; changing the law first lets `can_equip_artifact` pass. The effect only adds the law + sets a var (no dependency on equipping), so the reorder is side-effect-free. All effect blocks and their `show_as_tooltip` mirrors are reordered consistently; the unrelated `coronation_change_law_effect` in event 6103 (no equip) is correctly left alone. `make tiger` clean (0/0, equal to main).

PR #546, opened 2026-06-20
Use neutral suzerain wording for non-mandala tributary notification

### LINKED ISSUE #539: become_tributary notification uses mandala-only 'Devaraja' wording for non-mandala tributaries
**Reported on Steam by `Nider001` on `2026-06-19 10:20`** (https://steamcommunity.com/workshop/filedetails/discussion/2871648329/4697908845434539387/).

> The "become_tributary_interaction" interaction, which is generally available to all playable government rulers, uses the "mandala_end_tributary_with_notification_effect = yes" effect to notify the former suzerain (if there is one) about switching sides. The effect calls the "tgp_east_asia_interaction_events.0010" event to send a notification. However, this event fails to account for non-mandala tributaries, resulting in always sending the following text:
>
> tgp_east_asia_interaction_events.0010.desc: "I serve another Devaraja now. One that will show us the true way.\n\nAll hail [leaving_tributary.GetSuzerain.GetName]!"
> This makes no sense for non-mandala government types, such as feudal or tribal.
>
> The simplest workaround is to replace "Devaraja" with "suzerain" in the abovementioned localization string. However, this will also affect mandala rulers. Alternatively, modify the event description to have a custom localization string for non-mandala rulers, that mentions "suzerain".

pharaox:
Fixes #539

**fixer:** `become_tributary_interaction` is available to all governments and, when the actor is already a tributary, calls `mandala_end_tributary_with_notification_effect`, which fires `tgp_east_asia_interaction_events.0010` to the former suzerain regardless of government type. That event's `desc` hardcodes the mandala-only word "Devaraja", which reads wrong for feudal/tribal tributaries.

Added a `triggered_desc` on `.0010` (mirroring the existing `.0035` pattern in the same file) keyed on `scope:leaving_tributary` having the `government_is_mandala` flag: mandala tributaries keep the original "Devaraja" flavor, while everyone else gets a new neutral "suzerain" key. The vanilla event file is added unmodified in a separate commit; the new key lives in `unop_new_keys_l_english.yml` and the original mandala key is untouched.


events/dlc/tgp/tgp_east_asia_interaction_events.txt
```diff
@@ -0,0 +1,1115 @@
...
+
+#Notify the former Suzerain that they were dropped!
+tgp_east_asia_interaction_events.0010 = {
+	type = letter_event
+	opening = tgp_east_asia_interaction_events.0010.opening
+	#Unop Use the mandala-only "Devaraja" wording only for mandala tributaries
+	desc = {
+		first_valid = {
+			triggered_desc = {
+				trigger = { scope:leaving_tributary = { government_has_flag = government_is_mandala } }
```
pharaox, line 12: **reviewer:** Discriminator is sound: `.0010` has a single caller (`mandala_end_tributary_with_notification_effect`) which does `save_scope_as = leaving_tributary` right before triggering this event, so `scope:leaving_tributary` always exists here — no missing-scope risk — and it's also the event's `sender`, i.e. the voice speaking the letter, so keying the wording on its government is correct.

pharaox (commented):
**reviewer:** Looks good — government-conditional `triggered_desc` keyed on the sender (`leaving_tributary`) is the right approach: mandala keeps the "Devaraja" flavor, everyone else gets a neutral "suzerain" key (the analysis's preferred non-lossy fix). Mirrors the in-file `.0035` pattern; the discriminator scope is always saved by the sole caller before `.0010` fires; the new loc key is an exact mirror of vanilla with only the one word changed; commit 1 is byte-identical vanilla; `make tiger` clean. Only a subjective in-game glance at the wording remains.

PR #547, opened 2026-06-20
Stop hold_court church/city events building in baronies that can't have holdings

### LINKED ISSUE #532: hold_court.8060 and hold_court.8070 can build in baronies that can't have holdings (e.g. b_djado_route)
**Reported on Steam by `White Lich` on `2026-06-19 04:27`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> Two royal court events hold_court.8060 and hold_court.8070 do not check if the barony it is building in are allowed to have holdings and thus can build in invalid baronies like b_djado_route. There might be more, those are two that I have noticed thus far.

pharaox:
Fixes #532

**fixer:** hold_court church/city events (`hold_court.8060`/`8070`) could pick baronies that can't have holdings (e.g. `b_djado_route` in the Sahara corridor).

The fix lives in the shared trigger, not in the hold_court events:

- `common/scripted_triggers/02_ep1_scripted_triggers.txt`: added `barony_cannot_construct_holding = no` to `province_has_no_holding_trigger`. Because `county_has_empty_province_trigger` is defined as `any_county_province = { province_has_no_holding_trigger = yes }`, tightening the base trigger propagates to both the province pick and the county gate, so county-selection and province-selection stay consistent and hold_court 8060/8070 are fixed without editing them. Every other `= yes` build-site caller (bp2, laamp, petition_liege, tgp_mandala, the seasia mosque path) inherits the fix too. Vanilla `mpo_decisions.txt` pairs `has_holding = no` with `barony_cannot_construct_holding = no` at its build sites, corroborating the discriminator.
- The only inverted (`= no`) usage in all of vanilla is `pilgrimage_seasia.1000`, which picks a *developed* merchant home. Its meaning must be preserved, so the original trigger body is inlined there — now written as `OR = { has_ongoing_construction = yes  has_holding = yes }` (the De Morgan form, per review, avoiding a multi-entry `NOT`).
- Also fixes a genuine latent bug in the same newly-added seasia file: the mass-convert option ran `set_county_faith` on a province scope instead of `.county`, so the conversion silently no-op'd. It's incidental to #532 but a correct fix in a file this PR already adds — happy to split it into its own PR if preferred.

Vanilla seasia file added unmodified in its own commit; fix in follow-ups. `make tiger` clean (0/0).

### DIFF
--- a/common/scripted_triggers/02_ep1_scripted_triggers.txt
+++ b/common/scripted_triggers/02_ep1_scripted_triggers.txt
@@ -311,6 +311,7 @@ target_of_powerful_faction_trigger = {
 province_has_no_holding_trigger = { # province has no holding
 	has_ongoing_construction = no
 	has_holding = no
+	barony_cannot_construct_holding = no #Unop also require that a holding is actually allowed here (excludes e.g. Sahara corridor baronies)
 }
 
 county_has_all_holding_types = { # county has all three holding types
--- a/events/activities/pilgrimage_activity/pilgrimage_events_seasia.txt
+++ b/events/activities/pilgrimage_activity/pilgrimage_events_seasia.txt
@@ -1,0 +1,2758 @@
+﻿# Events for Pilgrimages in South-East Asia
+
+namespace = pilgrimage_seasia
+
+# The Pearl Diver
+pilgrimage_seasia.100 = {
+	type = character_event
+	title = pilgrimage_seasia.100.t
+	desc = pilgrimage_seasia.100.desc
+	theme = travel_pilgrimage
+	override_background = { reference = bp3_coast }
+	left_portrait = {
+		character = root
+		animation = interested
+	}
+	right_portrait = {
+		character = scope:pearl_diver_child
+		animation = storyteller
+		camera = camera_event_right_pointing_right
+	}
+	cooldown = { years = 10 }
+
+	trigger = {
+		has_tgp_dlc_trigger = yes
+		is_travelling = yes
+		OR = {
+			government_has_flag = government_is_mandala
+			religion = { is_in_family = rf_eastern }
+			faith = faith:aluk
+			faith = faith:dayawism
+			faith = faith:hantuism
+			faith = faith:kaharingan
+			faith = faith:satsana_phi
+			faith = faith:tolotang
+			faith = faith:utaki
+		}
+		is_adult = yes
+		location = {
+			has_sea_danger_type = { TRAVEL = root.current_travel_plan }
+			any_neighboring_province = {
+				exists = county
+				OR = {
+					geographical_region = world_asia_southeast_mainland
+					geographical_region = world_asia_southeast_islands
+				}
+				is_coastal = yes
+			}
+		}
+	}
+
+	immediate = {
+		mp_delay_travel_plan = { DAYS = 90 }
+		location = {
+			random_neighboring_province = {
+				limit = {
+					exists = county
+					OR = {
+						geographical_region = world_asia_southeast_mainland
+						geographical_region = world_asia_southeast_islands
+					}
+					is_coastal = yes
+				}
+				save_scope_as = background_terrain_scope
+			}
+		}
+		create_character = {
+			location = scope:background_terrain_scope
+			template = peasant_child_character
+			age = { 8 11 }
+			gender_female_chance = root_faith_dominant_gender_adjusted_female_chance
+			faith = scope:background_terrain_scope.faith
+			culture = scope:background_terrain_scope.culture
+			save_scope_as = pearl_diver_child
+		}
+	}
+
+	option = { # Adopt child
+		name = pilgrimage_seasia.100.a
+		
+		trigger = {
+			can_be_parent_of = scope:pearl_diver_child
+			has_activity_intent = altruism_intent
+		}
+		
+		reason = activity_intent
+		
+		very_pious_type_option_effect = yes
+		
+		scope:pearl_diver_child = {
+			add_piety_level = 4
+		}
+		
+		add_to_entourage_court_and_activity_effect = { 
+			CHAR_TO_ADD = scope:pearl_diver_child 
+			NEW_COURT_OWNER = root 
+		}
+		
+		if = {
+			limit = {
+				is_male = yes
+			}
+			scope:pearl_diver_child = { set_father = root }
+		}
+		else = {
+			scope:pearl_diver_child = { set_mother = root }
+		}
+		if = {
+			limit = { exists = house }
+			scope:pearl_diver_child = { set_house = root.house }
+		}
+		set_relation_friend = {
+			reason = friend_adopted
+			target = scope:pearl_diver_child
+		}
+		
+		stress_impact = {
+			sadistic = massive_stress_impact_gain
+			cynical = massive_stress_impact_gain
+			compassionate = massive_stress_impact_loss
+			zealous = medium_stress_impact_loss
+		}
+		ai_chance = {
+			base = 5
+			modifier = {
+				OR = {
+					has_trait = cynical
+					has_trait = sadistic
+				}
+				factor = 0
+			}
+		}
+	}
+
+	option = { # Accept pearl as an omen
+		name = pilgrimage_seasia.100.b
+		
+		add_piety = minor_piety_gain
+		pious_type_option_effect = yes
+		
+		stress_impact = {
+			cynical = medium_stress_impact_gain
+			zealous = minor_stress_impact_loss
+		}
+		ai_chance = {
+			base = 100
+			modifier = {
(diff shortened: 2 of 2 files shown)


pharaox:
**fixer:** Reworked per the maintainer's suggestion — much smaller diff.

Instead of adding sibling triggers and swapping ~30 hold_court filters, the guard now lives in the shared trigger:

- `02_ep1_scripted_triggers.txt`: added `barony_cannot_construct_holding = no` to `province_has_no_holding_trigger`. Every `= yes` caller (hold_court, court, petition_liege, the bp2/coronation/laamp/mpo/ep3 paths, etc.) inherits the fix automatically, and the wrappers `county_has_empty_province_trigger` / `county_can_*` tighten correctly too.
- Vetted **all** usages in mod + vanilla. The only one that isn't a "can a holding be built here" check is `pilgrimage_events_seasia.1000`, which uses `province_has_no_holding_trigger = no` to pick a *developed* coastal county as a merchant's home. To preserve its semantics I added that vanilla file and inlined the original trigger body there (`NOT = { has_ongoing_construction = no  has_holding = no }`).
- The mosque-build path in the same pilgrimage file (`= yes` → `begin_create_holding`) was a previously-missed at-risk caller and is now guarded by the shared change.
- Tiger surfaced one unrelated genuine vanilla bug in the newly-added file: the SE-Asia mass-convert option ran `set_county_faith` on a province scope (`scope:background_terrain_scope`) instead of `.county`, so the conversion silently no-op'd. Fixed in the same PR.

Assumption (corroborated): `barony_cannot_construct_holding = no` is valid in province scope — vanilla `00_decisions_effects.txt` already uses it inside `random_county_province { limit }` beside `has_holding = no`.

events/activities/pilgrimage_activity/pilgrimage_events_seasia.txt
```diff
@@ -1616,7 +1616,7 @@ pilgrimage_seasia.1000 = {
 				limit = {
 					tier = tier_county
 					religion = religion:islam_religion
-					title_province = { province_has_no_holding_trigger = no }
+					title_province = { NOT = { has_ongoing_construction = no  has_holding = no } } #Unop inline the former province_has_no_holding_trigger body: this picks a developed merchant home, unrelated to whether a holding can be built
```
pharaox: Format properly using the correct indentation

pharaox: **fixer:** Done — expanded the inlined `NOT { ... }` into a properly-indented multi-line block matching the surrounding code.

pharaox:
**fixer:** Reformatted the inlined trigger at `pilgrimage_seasia.1000` to proper multi-line indentation matching the surrounding `limit` block (the `NOT { has_ongoing_construction / has_holding }` is now expanded across lines, with the `#Unop` note on the first line). No logic change; `make tiger` clean.

common/scripted_triggers/02_ep1_scripted_triggers.txt
```diff
@@ -311,6 +311,7 @@ target_of_powerful_faction_trigger = {
 province_has_no_holding_trigger = { # province has no holding
 	has_ongoing_construction = no
 	has_holding = no
+	barony_cannot_construct_holding = no #Unop also require that a holding is actually allowed here (excludes e.g. Sahara corridor baronies)
```
pharaox, line 314: **reviewer:** This is the whole fix, and it's the right call. Because `county_has_empty_province_trigger` is defined as `any_county_province = { province_has_no_holding_trigger = yes }`, tightening the base trigger here propagates to both the province pick and the county gate — so the consistency the PR description worried about is guaranteed, and the hold_court 8060/8070 events are fixed without touching them. Vanilla `mpo_decisions.txt` pairs `has_holding = no` with `barony_cannot_construct_holding = no` at its build sites, which confirms the discriminator.

events/activities/pilgrimage_activity/pilgrimage_events_seasia.txt
```diff
@@ -0,0 +1,2764 @@
...
+		}
+		title:c_mecca.kingdom = {
+			random_in_de_jure_hierarchy = {
+				continue = {
+					tier >= tier_duchy
+				}
+				limit = {
+					tier = tier_county
+					religion = religion:islam_religion
+					#Unop inline the former province_has_no_holding_trigger body: this picks a developed merchant home, unrelated to whether a holding can be built
```
pharaox: **reviewer:** Good catch handling this — it's the only `province_has_no_holding_trigger = no` usage in all of vanilla, where the intent is the opposite (a *developed* merchant home). Inlining the old body keeps its meaning intact after the shared trigger was tightened.

events/activities/pilgrimage_activity/pilgrimage_events_seasia.txt
```diff
@@ -0,0 +1,2764 @@
...
+		
+		add_piety = minor_piety_gain
+		
+		pious_type_option_effect = yes
+		
+		if = {
+			limit = {
+				scope:background_terrain_scope.faith != root.faith
+			}
+			scope:background_terrain_scope.county = { #Unop set_county_faith needs a county scope, not a province
```
pharaox, line 790: **reviewer:** `Low` — this `set_county_faith` scope fix is correct (`scope:background_terrain_scope` is a province, not a county), but it's unrelated to #532. Consider splitting it into its own PR per the "minimum change / avoid incidental edits" guideline.

pharaox (commented):
**reviewer:** Looks good — the shipped fix is sound and actually a cleaner, higher-altitude change than the PR description (which is stale).

The real diff is one line: `barony_cannot_construct_holding = no` added to the shared `province_has_no_holding_trigger` — not the two new triggers / hold_court edits the description claims. That works well because `county_has_empty_province_trigger` is defined in terms of this trigger, so county-gating and province-pick stay consistent automatically, and every `= yes` caller (hold_court 8060/8070, bp2, laamp, petition_liege, tgp_mandala, seasia mosque) is a build site that gets fixed by the one change. Vanilla `mpo_decisions.txt` pairs `has_holding = no` + `barony_cannot_construct_holding = no` the same way, corroborating the discriminator. The only inverted (`= no`) usage in vanilla (seasia:1619) was correctly preserved by inlining the old body. Tiger clean, commit hygiene correct.

Two `Low` notes, neither blocking:
- The PR description describes a rejected approach — please update it to match the shared-trigger change.
- An unrelated `set_county_faith` scope fix is bundled in (seasia:790); correct, but incidental to #532.

events/activities/pilgrimage_activity/pilgrimage_events_seasia.txt
```diff
@@ -1616,7 +1616,13 @@ pilgrimage_seasia.1000 = {
 				limit = {
 					tier = tier_county
 					religion = religion:islam_religion
-					title_province = { province_has_no_holding_trigger = no }
+					#Unop inline the former province_has_no_holding_trigger body: this picks a developed merchant home, unrelated to whether a holding can be built
+					title_province = {
+						NOT = {
```
ProZeratul: NOT with multiple condition shouldn't be a thing, always use NOR or NAND

by default a NOT with multiple entries is a NOR, but it can be confusing when reviewing (and I know that even scripters get confused and think this would be a NAND)

ProZeratul (changes requested):

events/activities/pilgrimage_activity/pilgrimage_events_seasia.txt
pharaox: Good point. Also, NOR with 2 `no` is more readable as OR with 2 `yes` (De Morgan's law).

pharaox: **fixer:** Done. Rewrote the inlined check as `OR = { has_ongoing_construction = yes  has_holding = yes }` — the De Morgan equivalent of the old `NOT = { ... = no  ... = no }`, so no multi-entry `NOT`/implicit `NOR`, and it reads directly as "a developed province" (matching pharaox's suggestion). `make tiger` clean.

pharaox:
**fixer:** Addressed the changes-requested review:

- **Multi-entry `NOT` (seasia:1621):** rewritten as `OR = { has_ongoing_construction = yes  has_holding = yes }` — the De Morgan equivalent of the old `NOT = { ... = no  ... = no }`, so no implicit `NOR` and it reads directly as "a developed province". `make tiger` clean.
- **Stale PR description (Low):** updated the body to describe the actual shipped change (the single shared-trigger line), not the rejected two-trigger approach.
- **Incidental `set_county_faith` fix (Low):** it's a correct province→county scope fix in a file this PR already adds; I've left it in and called it out in the description. Say the word if you'd rather I split it into its own PR.

PR #548, opened 2026-06-20
Stop GOK Onslaught devastation razing silk road market holdings

### LINKED ISSUE #533: Greatest of All Khans Onslaught CB can raze the Changan silk road market building, breaking the silk road situation
**Reported on Steam by `White Lich` on `2026-06-19 07:36`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> The Greatest of All Khans Onslaught casus belli can raze the silk road market building in Changan. Which causes alot of weirdness in the silk road situation like perpetual hardship because catalysts now broken. I think that's the only important one they can destroy because all the other ones relevant to the situation are in the county capitals which they seem to be unable to destroy.

pharaox:
Fixes #533

**fixer:** The Greatest of All Khans Onslaught CB runs `gok_realm_devastation_effect` on victory. In the refused-submission branch, its `remove_holding` loop razes whole non-capital holdings in the devastation area (`has_holding = yes`, `is_county_capital = no`, 50% chance) with no protection for holdings carrying a special situation building — so the Changan silk road market (a non-capital `silk_road_node` special building) could be destroyed, breaking the silk road situation. Capital silk-road markets already survive via the existing `is_county_capital = no` guard.

Added `NOT = { has_building_with_flag = { flag = silk_road_node } }` to the per-province `remove_holding` limit so silk-road node holdings are never razed. The building-destruction path (`destroy_random_building_*`) already cannot select `type = special` buildings, so it needs no change.

### DIFF
--- a/common/scripted_effects/09_mpo_greatest_of_khans_effects.txt
+++ b/common/scripted_effects/09_mpo_greatest_of_khans_effects.txt
@@ -7062,6 +7062,7 @@ gok_realm_devastation_effect = {
 					limit = {
 						has_holding = yes
 						is_county_capital = no
+						NOT = { has_building_with_flag = { flag = silk_road_node } } #Unop Don't raze silk road market holdings, as that breaks the silk road situation
 					}
 					random = {
 						chance = 50


common/scripted_effects/09_mpo_greatest_of_khans_effects.txt
```diff
@@ -7062,6 +7062,7 @@ gok_realm_devastation_effect = {
 					limit = {
 						has_holding = yes
 						is_county_capital = no
+						NOT = { has_building_with_flag = { flag = silk_road_node } } #Unop Don't raze silk road market holdings, as that breaks the silk road situation
```
pharaox, line 7065: **reviewer:** Correct scope and placement — `has_building_with_flag` is valid in this province scope, and putting it in the inner `every_county_province` limit (not the outer county limit) means only the silk-road province is protected while the rest of the devastation area still razes as intended. This is the only `remove_holding` in the mod, so the silk-road market can no longer be destroyed by the Onslaught CB.

pharaox (commented):
**reviewer:** Looks good — the guard sits in the inner per-province `limit` beside the existing `is_county_capital = no` exclusion, so the sole `remove_holding` in `gok_realm_devastation_effect` now skips silk-road-node holdings. Minimal and correct: leaving the outer county `limit` alone still lets other non-capital provinces be razed, only the silk-road province is spared (no rebalance). `silk_road_node` is the right discriminator (set on the market specials), the block-form trigger is a vanilla idiom, the `destroy_random_building_*` path can't hit `type = special`, and `make tiger` is clean.

PR #550, opened 2026-06-25
fix syntax in conqueror events

Svatopluk1:
Fix lots of trigger syntax issues.

### DIFF
--- a/events/story_cycles/story_cycle_conqueror_events.txt
+++ b/events/story_cycles/story_cycle_conqueror_events.txt
@@ -1,4 +1,4 @@
-﻿
+
 namespace = conqueror
 
 # Become a Conqueror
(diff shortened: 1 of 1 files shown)


Nider001:
The text file is in UTF-8 format rather than UTF-8-BOM, causing a warning in logs

Svatopluk1:
Hopefully the text file shows up as UTF-8-BOM now.
Also added #unop comments to the changed/added lines.

ProZeratul:
@Svatopluk1 The `vassal_stance` are valid script and run fine in-game, you can check this in the debugger in-game. Same observation apply for `type = story_conqueror`

Nonetheless You were right on the parochial filtering missing, so I've rechecked orther files and it was also missing here & there so good catch on that one.

I'll close this PR as the parochial filtering have been made in https://github.com/ProZeratul/CrusaderKings3UnofficialPatch/pull/564

PR #569, opened 2026-06-28
Fix subjugated tributary expected/rich collapse and false guarantee tooltip

### LINKED ISSUE #540: Tributary 'expected' vs 'high' tribute identical for non-mandala governments
**Reported on Steam by `Captain Smollett` on `2026-06-19 12:28`** (https://steamcommunity.com/workshop/filedetails/discussion/2871648329/4697908845434539387/).

> About tributaries, i'll add that due to how the start_tributary_interaction_effect is setup, for tribal, feudal and clan (and maybe some other ones?) governments there's no difference between submitting to suzerain with "expected tribute" or "high tribute" other than acceptance and opinion gain, which is weird.

pharaox:
Fixes #540

**fixer:** The subjugated branch of `start_tributary_interaction_effect` had only two obligation tiers (`if low / else`), so the "expected" and "rich" send-options produced identical contract obligations, and it emitted the `suzerain_guarantee` tooltip unconditionally even though the meager tier never grants the guarantee.

Restructured the branch into the same three-tier `if high / else_if low / else normal` shape used by every sibling government block:
- **rich (high):** taxes 2, prestige **2** (`prestige_transfer_high`) + guarantee — now distinct from expected (taxes already maxed at 2, so prestige is the differentiator, matching the steppe/nomad precedent).
- **expected (normal):** unchanged — taxes 2, prestige 1, guarantee.
- **meager (low):** taxes 1, prestige 0, guarantee explicitly set to none and tooltip corrected to `no_suzerain_guarantee`.

Adds new loc key `high_subjugated_tribute_prestige` (mirrors the vanilla none/normal subjugated prestige keys). `make tiger` clean.

### DIFF
--- a/common/scripted_effects/00_interaction_effects.txt
+++ b/common/scripted_effects/00_interaction_effects.txt
@@ -2912,17 +2912,40 @@ start_tributary_interaction_effect = {
 				contract_group = tributary_subjugated
 				suzerain = $SUZERAIN$
 			}
-			if = {
+			if = { #Unop give the rich (high) tier a distinct, higher prestige obligation so expected != rich
+				limit = {
+					OR = {
+						scope:high_obligations ?= yes
+						exists = scope:tributary_war
+					}
+				}
+				if = {
+					limit = { is_tributary = yes } # suppress error logging
+					hidden_effect = {
+						tributary_contract_set_obligation_level = { type = default_tributary_taxes level = 2 }
+						tributary_contract_set_obligation_level = { type = default_tributary_prestige level = 2 }
+						tributary_contract_set_obligation_level = { type = suzerain_war_participation_guarantee level = 1 }
+						tributary_contract_set_obligation_level = { type = tributary_war_participation_obligation level = 0 }
+					}
+				}
+				custom_tooltip = high_subjugated_tribute_taxes
+				custom_tooltip = high_subjugated_tribute_prestige
+				custom_tooltip = suzerain_guarantee
+			}
+			else_if = {
 				limit = { scope:low_obligations ?= yes }
 				if = {
 					limit = { is_tributary = yes } # suppress error logging
 					hidden_effect = {
 						tributary_contract_set_obligation_level = { type = default_tributary_taxes level = 1 }
 						tributary_contract_set_obligation_level = { type = default_tributary_prestige level = 0 }
+						tributary_contract_set_obligation_level = { type = suzerain_war_participation_guarantee level = 0 } #Unop meager tier does not grant the guarantee
+						tributary_contract_set_obligation_level = { type = tributary_war_participation_obligation level = 0 }
 					}
 				}
 				custom_tooltip = low_subjugated_tribute_taxes
 				custom_tooltip = none_subjugated_tribute_prestige
+				custom_tooltip = no_suzerain_guarantee #Unop tooltip now matches the obligation actually set
 			}
 			else = {
 				if = {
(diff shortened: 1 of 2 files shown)


common/scripted_effects/00_interaction_effects.txt
```diff
@@ -2936,8 +2959,8 @@ start_tributary_interaction_effect = {
 				}
 				custom_tooltip = high_subjugated_tribute_taxes
 				custom_tooltip = normal_subjugated_tribute_prestige
+				custom_tooltip = suzerain_guarantee
```
pharaox: **reviewer:** This `custom_tooltip = suzerain_guarantee` was relocated here from the old unconditional position, but the changed line carries no `#Unop` marker. Behavior is correct; just add a marker for consistency with the style guide.

```suggestion
				custom_tooltip = suzerain_guarantee #Unop tooltip now scoped to the normal tier
```

pharaox (commented):
**reviewer:** Looks good — the subjugated branch is restructured into the canonical 3-tier high/low/normal shape used by every sibling government block, and both reported bugs are fixed: rich now sets prestige level 2 (distinct from expected's level 1), and the meager tier sets the guarantee to level 0 with a matching `no_suzerain_guarantee` tooltip. Verified the obligation types/levels and all tooltip keys exist in vanilla; the new `high_subjugated_tribute_prestige` loc key mirrors the vanilla none/normal keys. War-subjugation now lands in the high tier (prestige=2), consistent with every sibling. `make tiger` re-run clean.

- Low: the relocated `custom_tooltip = suzerain_guarantee` added into the normal block lacks a `#Unop` marker.

Nider001:
I still stand by my original proposal regarding the new tribute levels, written in the issue comment, as I believe they would make more sense than what the system just came up with. As in, add Suzerain Guarantee to meager tribute (to match the vanilla tooltip) and remove Suzerain Guarantee from rich tribute.

Firstly, why would meager (tier 1) tribute have no guarantee when expected (tier 2) tribute has it? This breaks the tier hierarchy. To illustrate the point, it's like if low feudal taxes gave more gold to liege than normal/high taxes. Not to mention the vanilla tooltip stating Suzerain Guarantee for meager tribute.

Secondly, with current implementation the only difference between the newly added rich tribute and the old expected tribute is slightly more prestige transfer - this is too little to justify a new tier. The rich tribute level is also enforced upon winning a tributary war, so it makes perfect sense to have no Suzerain Guarantee on top of higher prestige transfer.

pharaox:
@Nider001 No need to post the same comment on the issue and the PR, this can only confuse our AI agents. Use the issue for discussions, unless you have concrete comments on the code itself.

pharaox:
Converting back to Draft as there is an ongoing discussion, please don't merge.

pharaox:
**fixer:** Reworked per the latest discussion (@pharaox @Nider001):

- **Subjugated:** tiers set to meager 1t/0p, expected 2t/1p, rich 2t/2p. Suzerain Guarantee is now granted to **all three** tiers — moved the `suzerain_war_participation_guarantee`/`tributary_war_participation_obligation` obligations out of the per-tier `if` so they always apply, matching the unconditional `suzerain_guarantee` tooltip and the other tributary types. Only the added rich tier (prestige level 2) differs from vanilla; the false-guarantee tooltip is resolved by making the obligation match the tooltip rather than the reverse.
- **Wanua:** meager prestige nudged from level 1 to level 0, so meager (barter 1 / tax 1 / prestige 0) is now distinct from expected (barter 2 / tax 1 / prestige 1) and matches its `none_subjugated_tribute_prestige` tooltip.
- **Steppe:** setup tooltips fixed — the block sets `default_tributary_taxes` but showed `*_nomad_tribute_herd` keys. Now shows `high_/low_subjugated_tribute_taxes` matching the actual obligation (normal and rich both set tax level 2 → both show `high_subjugated_tribute_taxes`; taxes cap at level 2 for this contract).

`make tiger` clean (0/0/0). Also rebased onto `main` to pick up the recently-merged goryeo loc key, which the branch would otherwise have dropped.

PR #571, opened 2026-06-28
Allow recreating e_goryeo after destruction once formation decision is consumed

### LINKED ISSUE #556: Samhan empire cannot be recreated after destruction (locked behind Unite the Husamguk decision)
**Reported on Steam by `Nider001` on `2026-06-25 15:36`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> There's an issue similar to the one with the Roman hegemony and Germania empire title creation you resolved previously. In my game, a Silla culture (Korean) ruler used the "Unite the Husamguk" decision to create the Samhan empire, then eventually took over Hegemony of China (yes, that can happen). As a result, the Samhan empire got destroyed and can no longer be created, because it is locked behind the decision, which is no longer available. In fact, it seems the empire can not be recreated by Samhanic heritage rulers after being destroyed for any reason.

pharaox:
Fixes #556

**fixer:** `e_goryeo` (Samhan/Goryeo) can only be formed by the one-time `tgp_korea_unify_goryeo_decision`, which permanently removes itself via `unavailable_unique_decisions`. Its `can_create` then blocks the intended formers — a Korean/Samhanic **administrative** ruler — from creating it directly (`NOT = { government_allows = administrative }`). So once the empire is destroyed (e.g. by taking Hegemony of China, as reported), there is no path left to recreate it.

Mirroring the #440 / Germania precedent, this relaxes `e_goryeo.can_create`: an administrative Korean-heritage ruler may now create `e_goryeo` directly **only after** the formation decision is no longer available *and* the title currently has no holder — i.e. exactly the post-destruction lockout. The normal pre-formation flow (use the decision) and its one-time rewards are untouched, and a second concurrent `e_goryeo` is still prevented by the `NOT = { exists = title:e_goryeo.holder }` guard.

Adds vanilla `05_goryeo.txt` (unmodified, separate commit) and a new loc key `unify_goryeo_recreate_after_destruction_tt`. `make tiger` clean.

### DIFF
--- a/common/landed_titles/05_goryeo.txt
+++ b/common/landed_titles/05_goryeo.txt
@@ -18,9 +18,19 @@ e_goryeo = {
 					}
 				}
 			}
-			custom_tooltip = {
-				text = unify_goryeo_decision_requirement_tt
-				NOT = { government_allows = administrative }
+			OR = { #Unop also allow recreation after e_goryeo is destroyed, once the one-time formation decision is gone
+				custom_tooltip = {
+					text = unify_goryeo_decision_requirement_tt
+					NOT = { government_allows = administrative }
+				}
+				custom_description = {
+					text = unify_goryeo_recreate_after_destruction_tt
+					NOT = { exists = title:e_goryeo.holder }
+					is_target_in_global_variable_list = {
+						name = unavailable_unique_decisions
+						target = flag:unify_goryeo_decision
+					}
+				}
 			}
 		}
 	}
--- a/common/landed_titles/05_goryeo.txt
+++ b/common/landed_titles/05_goryeo.txt
@@ -1,0 +1,1176 @@
+﻿@correct_culture_primary_score = 100
+@better_than_the_alternatives_score = 50
+@always_primary_score = 1000
+@never_primary_score = -1000
+
+e_goryeo = {
+	color = { 220 205 250 }
+
+	capital = c_gaeseong
+
+	can_create = {
+		trigger_if = {
+			limit = {
+				culture = {
+					OR = {
+						has_cultural_pillar = heritage_korean
+						any_parent_culture = { has_cultural_pillar = heritage_korean }
+					}
+				}
+			}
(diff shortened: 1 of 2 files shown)


common/landed_titles/05_goryeo.txt
```diff
@@ -0,0 +1,1186 @@
...
+			limit = {
+				culture = {
+					OR = {
+						has_cultural_pillar = heritage_korean
+						any_parent_culture = { has_cultural_pillar = heritage_korean }
+					}
+				}
+			}
+			OR = { #Unop also allow recreation after e_goryeo is destroyed, once the one-time formation decision is gone
+				custom_tooltip = {
```
pharaox, line 22: **reviewer:** Correct hinge. `is_target_in_global_variable_list { unavailable_unique_decisions / flag:unify_goryeo_decision }` is true exactly after `tgp_korea_unify_goryeo_decision`'s `effect` runs once, so this recreate branch can't be used to bypass the decision pre-formation, and `NOT = exists title:e_goryeo.holder` blocks a concurrent second `e_goryeo`. Matches the #440 design intent.

pharaox (commented):
**reviewer:** Looks good — relaxes `e_goryeo.can_create` to re-admit an administrative Korean ruler only after the one-time `unify_goryeo_decision` is consumed and the title has no holder, mirroring the #440 precedent. Verified the recreate branch keys on the exact flag the decision's `effect` sets, so the path opens only post-formation (pre-formation admins stay blocked); the `heritage_korean` `trigger_if` limit and `NOT exists e_goryeo.holder` leave all other cases unchanged and prevent a concurrent second empire. Added `05_goryeo.txt` is byte-identical to vanilla in its own commit; `make tiger` clean on re-run.

- Low: sibling TGP formation empires (e.g. Japanese) may share the destruction-lockout idiom; out of scope here and deferred to a separate sweep — not blocking.

PR #572, opened 2026-06-28
Cap challenge_status influence transfer at the target's actual influence

### LINKED ISSUE #557: challenge_status scheme gives fixed 145 influence regardless of target, incentivizing bullying the weak
**Reported on Steam by `jy` on `2026-06-25 06:08`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> challenge_status scheme works for random titleless admin character with negative influence as target & always give a fixed value of 145, you get the same influence as targeting the emperor, it just incentivise bullying the weak
>
> in comparison, stealing herd gives % herd of the target + chance of backfiring
>
> propose fix
>
> 1, reward should be % influence of target (capped to a absolute value)
>
> 2, limit target to titled individuals
>
> point 2 might be too drastic a change, but without which, player can just invite an titleless influential character (e.g. former emperor) to their court, then challenge status, since scheme resistance is determine by their liege
>
> also, steal herd is limited to nomadic rulers, and herd is way more important to nomads than influence to admin rulers

pharaox:
Fixes #557

**fixer:** On a successful `challenge_status` scheme, `scheme_critical_moments.1172` sets the influence transfer to a flat `50 + scheme_success_chance` (~145 at max) with **no reference to the target** — so bullying a titleless pauper pays exactly the same as challenging the emperor, and the owner is credited the full amount even when the target doesn't have that much influence (influence conjured from nothing).

I treated this as a **correctness** fix rather than a balance rework (Unop fixes, it doesn't rebalance): the transfer is now clamped to the target's actual influence with `max = influence` (and floored at `min = 0`). The existing `50 + success_chance` remains the *maximum* payout, so strong targets are unaffected, but a weak target can only ever yield what they actually own — making the reward target-dependent and ensuring a true transfer. This mirrors the steal-herd scheme the reporter cites, which transfers a fraction of the target's actual resource.

I deliberately did **not** implement the reporter's other suggestions (a bespoke influence-scaling curve, or restricting targets to titled characters) — those are balance/design changes, flagged for maintainer judgment in the handover. `make tiger` clean.

### DIFF
--- a/events/scheme_events/scheme_critical_moments_events.txt
+++ b/events/scheme_events/scheme_critical_moments_events.txt
@@ -2461,13 +2461,13 @@ scheme_critical_moments.1172 = {
 		scope:target = {
 			save_scope_as = recipient # To fix loc
 			set_variable = {
-				#Unop Reward is a fraction of the target's own influence (steal_herd model), not the flat "50 + scheme_success_chance" which tied the payout to the unrelated success percentage and paid the same for a pauper as for the emperor
+				#Unop Reward is a fraction of the target's own influence, not the flat "50 + scheme_success_chance" which tied the payout to the unrelated success percentage and paid the same for a pauper as for the emperor
 				name = challenge_status_influence_loss
 				value = {
-					value = influence #Unop the target's current influence (this block runs inside scope:target)
-					multiply = 0.25 #Unop fraction of the target's influence transferred; magnitude is a balance call, flagged for tuning
-					min = 0 #Unop a target with no (or negative) influence yields nothing rather than conjuring a reward
-					max = 200 #Unop absolute ceiling; higher than the old ~145 max so rare 800+ influence targets compensate for the smaller reward on less influential ones
+					value = influence
+					multiply = 0.25
+					min = 0
+					max = 200
 				}
 			}
 		}
--- a/events/scheme_events/scheme_critical_moments_events.txt
+++ b/events/scheme_events/scheme_critical_moments_events.txt
@@ -2461,10 +2461,13 @@ scheme_critical_moments.1172 = {
 		scope:target = {
 			save_scope_as = recipient # To fix loc
 			set_variable = {
+				#Unop Reward is a fraction of the target's own influence (steal_herd model), not the flat "50 + scheme_success_chance" which tied the payout to the unrelated success percentage and paid the same for a pauper as for the emperor
 				name = challenge_status_influence_loss
 				value = {
-					value = 50
-					add = scope:scheme.scheme_success_chance
+					value = influence #Unop the target's current influence (this block runs inside scope:target)
+					multiply = 0.25 #Unop fraction of the target's influence transferred; magnitude is a balance call, flagged for tuning
+					min = 0 #Unop a target with no (or negative) influence yields nothing rather than conjuring a reward
+					max = 200 #Unop absolute ceiling; higher than the old ~145 max so rare 800+ influence targets compensate for the smaller reward on less influential ones
 				}
 			}
 		}


events/scheme_events/scheme_critical_moments_events.txt
```diff
@@ -2465,6 +2465,8 @@ scheme_critical_moments.1172 = {
 				value = {
 					value = 50
 					add = scope:scheme.scheme_success_chance
+					min = 0 #Unop a target with no influence can't pay out a reverse transfer
+					max = influence #Unop cap the transfer at the target's actual influence, so the reward scales with the target instead of being a flat amount conjured from nothing
```
pharaox: **reviewer:** Correct: this `set_variable` is inside `scope:target = { ... }`, so the bare `influence` resolves to the target's, and `min`/`max` clamp sequentially after `value`+`add`. Since this is the only write of `challenge_status_influence_loss` and 1174's real transfer reads it back, the clamp bounds both the target's loss and the owner's gain.

pharaox (commented):
**reviewer:** Looks good — clamps the influence transfer to the target's actual influence (`min = 0`, `max = influence`), making the reward target-dependent instead of a flat ~145. Verified the `set_variable` runs inside `scope:target = { ... }` so `influence` reads the target's; it's the sole computation site and every downstream use reads the stored var, so both the target's loss and the owner's gain are bounded by what the target has (a true transfer). The real effect lands in event 1174's immediate; 1172's option is display-only `show_as_tooltip`, so no double-application. Sound scoping under fix-not-rebalance — a bespoke scaling curve / titled-target restriction is left as a maintainer call. `make tiger` clean.

pharaox:
This requires another look, don't merge.

pharaox:
**fixer:** Reworked per the maintainer's call: the real bug is the reward formula conflating success chance with reward magnitude (`50 + scheme_success_chance` — a 95% success rate should not mean 145 influence). The previous clamp kept that formula and still paid ~145 against a rich target, so it's superseded.

New approach mirrors the `steal_herd` scheme (reward = a fraction of the *target's* own resource, capped), which the reporter cited as the consistent model:

```
value = influence      # target's current influence (block runs inside scope:target)
multiply = 0.25        # fraction transferred
min = 0                # a pauper (or negative influence) yields nothing
max = 145              # = old formula's max (50 + 95); never exceeds vanilla
```

This makes the reward genuinely target-dependent and independent of success chance, while the `max = 145` ceiling guarantees it never pays out more than vanilla could — so it's a correctness/anti-exploit fix, not a buff.

**Two knobs left as balance calls, flagged for tuning:** the fraction (`0.25`) and the absolute cap (`145`). I picked the cap to equal the old maximum so the change is strictly non-inflationary; adjust to taste. The reporter's second suggestion (restrict targets to titled characters) remains optional — scaling by target influence already removes the bully-the-weak incentive. `make tiger` clean.

pharaox:
`@fixer` Let's change the cap to 200. Targets with 800+ influence are rare, and having a higher rewqrd than 145 for them would compensate the lower reward on less influential targets.

events/scheme_events/scheme_critical_moments_events.txt
```diff
@@ -2461,10 +2461,13 @@ scheme_critical_moments.1172 = {
...
 			set_variable = {
+				#Unop Reward is a fraction of the target's own influence (steal_herd model), not the flat "50 + scheme_success_chance" which tied the payout to the unrelated success percentage and paid the same for a pauper as for the emperor
 				name = challenge_status_influence_loss
 				value = {
-					value = 50
-					add = scope:scheme.scheme_success_chance
+					value = influence #Unop the target's current influence (this block runs inside scope:target)
+					multiply = 0.25 #Unop fraction of the target's influence transferred; magnitude is a balance call, flagged for tuning
+					min = 0 #Unop a target with no (or negative) influence yields nothing rather than conjuring a reward
+					max = 145 #Unop absolute ceiling = the old formula's maximum (50 + 95), so this never pays out more than vanilla could
```
pharaox:
```suggestion
					value = influence
					multiply = 0.25
					min = 0
					max = 200
```

events/scheme_events/scheme_critical_moments_events.txt
```diff
@@ -2461,10 +2461,13 @@ scheme_critical_moments.1172 = {
 		scope:target = {
 			save_scope_as = recipient # To fix loc
 			set_variable = {
+				#Unop Reward is a fraction of the target's own influence (steal_herd model), not the flat "50 + scheme_success_chance" which tied the payout to the unrelated success percentage and paid the same for a pauper as for the emperor
```
pharaox:
```suggestion
				#Unop Reward is a fraction of the target's own influence, not the flat "50 + scheme_success_chance" which tied the payout to the unrelated success percentage and paid the same for a pauper as for the emperor
```

pharaox:
**fixer:** Raised the absolute cap from `145` to `200` per the maintainer's call. Rationale: targets with 800+ influence are rare, so letting the reward exceed the old ~145 max for them compensates for the smaller payout on less influential targets. The rest of the fix is unchanged — reward is `target.influence × 0.25`, floored at 0 (steal_herd model, independent of success chance). `make tiger` clean.

events/scheme_events/scheme_critical_moments_events.txt
```diff
@@ -2461,10 +2461,13 @@ scheme_critical_moments.1172 = {
...
 				name = challenge_status_influence_loss
 				value = {
-					value = 50
-					add = scope:scheme.scheme_success_chance
+					value = influence #Unop the target's current influence (this block runs inside scope:target)
+					multiply = 0.25 #Unop fraction of the target's influence transferred; magnitude is a balance call, flagged for tuning
+					min = 0 #Unop a target with no (or negative) influence yields nothing rather than conjuring a reward
+					max = 200 #Unop absolute ceiling; higher than the old ~145 max so rare 800+ influence targets compensate for the smaller reward on less influential ones
```
pharaox:
```suggestion
					value = influence
					multiply = 0.25
					min = 0
					max = 200
```

PR #573, opened 2026-06-28
Block Demand Submission or Ruin when target is imprisoned by the actor

### LINKED ISSUE #559: Demand Submission or Ruin (GOK) not blocked when target is imprisoned by the attacker
**Reported on Steam by `jy` on `2026-06-20 12:17`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> i been abusing this bug to great effect, the demand submission or ruin interaction of GOK doesnt get blocked by target being imprisoned by your character, i.e. you will immediately win the war (if target rejected offer) since war leader is already in prison

pharaox:
Fixes #559

**fixer:** The GOK "Offer Submission or Ruin" interaction had no guard against the recipient being the actor's own prisoner. `mpo_offer_submission_or_ruin_valid_recipient_trigger` only checked `is_independent_ruler`, `NOT is_at_war_with`, and the recent-independence/recent-offer flags — so you could demand submission from someone you already hold captive. On rejection the resulting subjugation war starts with the defender's war leader already imprisoned by the attacker → instant win, the reported exploit.

Added `NOT = { is_imprisoned_by = scope:actor }` to the recipient trigger, matching the established vanilla idiom (`00_war.txt` declare-war guards the recipient with `is_imprisoned + NOT imprisoner = scope:actor`; `00_gift.txt`, courtier/guest, revoke-title, diarch, ep3 interactions all use `NOT = { is_imprisoned_by = scope:actor }`). `is_imprisoned_by` carries built-in trigger localization, so the failure tooltip is auto-generated.

Placement: overrides `09_mpo_greatest_of_khans_triggers.txt` (added unmodified first) to keep recipient-validity logic where vanilla put it.

**Tiger:** adding the trigger file surfaced one pre-existing, unrelated vanilla scope warning — `mpo_war_of_defiance_notified_player_trigger` uses `root` as a character but is called from a casus belli `on_victory` block in `09_mpo_wars.txt`. That's a separate vanilla bug; ignored via a precise `ck3-tiger.conf` rule (mirroring the existing frankokratia-faith ignore) and flagged for separate follow-up. `make tiger` clean.

### DIFF
--- a/ck3-tiger.conf
+++ b/ck3-tiger.conf
@@ -597,11 +597,6 @@ filter = {
 			text = "expects root to be character but root seems to be faith"
 			file = common/scripted_triggers/07_frankokratia_triggers.txt
 		}
-		NAND = { # Pre-existing vanilla bug: mpo_war_of_defiance_notified_player_trigger uses root as a character but is called from a casus belli on_victory block (09_mpo_wars.txt); a separate fix, out of scope here
-			key = scopes
-			text = "expects root to be character but root seems to be casus belli"
-			file = common/scripted_triggers/09_mpo_greatest_of_khans_triggers.txt
-		}
 		NAND = { # Ignore scope warnings in event backgrounds, too difficult to fix
 			key = scopes
 			OR = {
@@ -652,6 +647,11 @@ filter = {
 			key = missing-item
 			text = "e_mongolia has no title history"
 			file = history/titles/k_dzungaria.txt
+		}
+		NAND = { # Pre-existing vanilla data quirk: attacker_wargoal_percentage = 1.5 (>1) in a Domination war; unrelated to this fix, clamping it would change war pacing - own issue
+			key = range
+			text = "should be between 0 (inclusive) and 1 (inclusive)"
+			file = common/casus_belli_types/09_mpo_wars.txt
 		}
 	}
 }
--- a/ck3-tiger.conf
+++ b/ck3-tiger.conf
@@ -648,7 +648,7 @@ filter = {
 			text = "e_mongolia has no title history"
 			file = history/titles/k_dzungaria.txt
 		}
-		NAND = { # Pre-existing vanilla data quirk: attacker_wargoal_percentage = 1.5 (>1) in a Domination war; unrelated to this fix, clamping it would change war pacing - own issue
+		NAND = { # Intended vanilla design: attacker_wargoal_percentage = 1.5 (>1) makes the attacker never tick warscore from occupation; vanilla uses the same value for nomadic conquest CBs (00_nomadic_conquest.txt), so clamping it would change intended war pacing
 			key = range
 			text = "should be between 0 (inclusive) and 1 (inclusive)"
 			file = common/casus_belli_types/09_mpo_wars.txt
--- a/ck3-tiger.conf
+++ b/ck3-tiger.conf
@@ -597,6 +597,11 @@ filter = {
 			text = "expects root to be character but root seems to be faith"
 			file = common/scripted_triggers/07_frankokratia_triggers.txt
 		}
+		NAND = { # Pre-existing vanilla bug: mpo_war_of_defiance_notified_player_trigger uses root as a character but is called from a casus belli on_victory block (09_mpo_wars.txt); a separate fix, out of scope here
+			key = scopes
+			text = "expects root to be character but root seems to be casus belli"
+			file = common/scripted_triggers/09_mpo_greatest_of_khans_triggers.txt
+		}
 		NAND = { # Ignore scope warnings in event backgrounds, too difficult to fix
 			key = scopes
 			OR = {
--- a/common/scripted_triggers/09_mpo_greatest_of_khans_triggers.txt
+++ b/common/scripted_triggers/09_mpo_greatest_of_khans_triggers.txt
@@ -239,50 +239,6 @@ mpo_war_of_defiance_notified_player_trigger = {
 	}
 }
 
-unop_war_of_defiance_notified_player_trigger = {
-	#Unop counterpart taking the gok as a parameter; the caller is a casus belli on_victory where root is the casus belli, not a character
-	NOT = {
-		this = $GOK$
-	}
-	OR = {
-		top_liege = $GOK$
-		top_suzerain = $GOK$
-		any_character_situation = {
-			this = situation:the_great_steppe
-		}
-		any_relation = {
-			type = rival
-			this = $GOK$
-		}
-		#Those within same de jure empire
-		capital_county ?= {
-			empire = {
-				any_de_jure_county = {
-					holder.top_liege = {
-						this = $GOK$
-					}
-				}
-			}
-		}
-		capital_county ?= {
-			empire = {
-				any_de_jure_county = {
-					holder.top_suzerain = {
-						this = $GOK$
-					}
-				}
-			}
-		}
-		AND = {
-			exists = $GOK$.capital_county
-			realm_to_title_distance_squared = {
-				target = $GOK$.capital_county
-				value <= squared_distance_almost_massive
-			}
-		}
-	}
-}
-
 gok_desirable_vassal_trigger = {
 	OR = {
 		ai_boldness <= low_negative_ai_value
--- a/common/scripted_triggers/09_mpo_greatest_of_khans_triggers.txt
+++ b/common/scripted_triggers/09_mpo_greatest_of_khans_triggers.txt
@@ -197,25 +197,26 @@ mpo_gok_submitting_coward_trigger = {
 }
 
 mpo_war_of_defiance_notified_player_trigger = {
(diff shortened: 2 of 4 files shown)


pharaox:
`@fixer` Fix the "pre-existing vanilla bug" reported by tiger. Also shorten the verbose `#Unop` comment.

common/scripted_triggers/09_mpo_greatest_of_khans_triggers.txt
```diff
@@ -0,0 +1,504 @@
...
+			}
+		}
+	}
+}
+mpo_offer_submission_or_ruin_valid_recipient_trigger = {
+	is_independent_ruler = yes
+	NOT = {
+		is_at_war_with = scope:actor
+	}
+	NOT = { is_imprisoned_by = scope:actor } #Unop can't demand submission from someone you already hold captive - on rejection the war's defender leader would already be imprisoned, an instant win
```
pharaox: **reviewer:** Good placement — this shared trigger is also called from the AI invasion target-selection in `09_mpo_greatest_of_khans_effects.txt` and from the interaction's `is_shown`/`is_valid_showing_failures_only`, so the guard covers the AI path too. `scope:actor` is already referenced here (`is_at_war_with = scope:actor`), so this introduces no new scope dependency.

ck3-tiger.conf
```diff
@@ -597,6 +597,11 @@ filter = {
 			text = "expects root to be character but root seems to be faith"
 			file = common/scripted_triggers/07_frankokratia_triggers.txt
 		}
```
pharaox, line 599: **reviewer:** Verified this is a real, pre-existing vanilla scope bug in `09_mpo_wars.txt:4561` (`mpo_war_of_defiance_notified_player_trigger` called with `root` = casus belli), not a false positive — so `# Ignored` is the correct section and deferring it (own file, own analysis) is the right call. Precise key+text+file rule, well grouped.

pharaox (commented):
**reviewer:** Looks good — adds `NOT = { is_imprisoned_by = scope:actor }` to the shared recipient-validity trigger, blocking the exact exploit (rejection-war defender leader already imprisoned → instant win), matching the `00_vassal_interactions.txt` precedent. Placement in the trigger also covers the AI Mongol-invasion target-selection sites in `09_mpo_greatest_of_khans_effects.txt`, not just the UI — a plus, and `scope:actor` was already used there so no new scope error (tiger clean). Vanilla file added byte-identical in its own commit.

Independently verified the tiger ignore: the `mpo_war_of_defiance_notified_player_trigger ... root seems to be casus belli` warning is a genuine pre-existing vanilla bug in `09_mpo_wars.txt:4561` (CB `on_victory` loop), unrelated to #559 and requiring a separate `09_mpo_wars.txt` override — correct to defer, correctly under `# Ignored` and grouped with the frankokratia scope ignore.

- Low: that deferred CB scope bug deserves its own follow-up issue, as flagged — not blocking.

pharaox:
**fixer:** Addressed both `@fixer` requests.

**Fixed the tiger-reported scope bug instead of ignoring it.** The root cause: in `09_mpo_wars.txt`'s `on_victory`, the `every_player` notification limit tried `scope:defender = { save_temporary_scope_as = root }` to make `root` the new Gurkhan, then called `mpo_war_of_defiance_notified_player_trigger`. That's an effect inside a trigger `OR` (a no-op that just returns true), and `root` in a CB `on_victory` is the casus belli — so the trigger's connection checks (`top_liege = root`, `root.capital_county`, …) ran against the CB. Net effect in vanilla: the OR short-circuits true and the event goes to *every* player.

Fix: parameterized `mpo_war_of_defiance_notified_player_trigger` with a `$GOK$` argument (replacing the broken `root`), dropped the no-op `save_temporary_scope_as` line, and call it as `mpo_war_of_defiance_notified_player_trigger = { GOK = scope:defender }`. Now only players actually connected to the new Gurkhan (vassal/suzerain, rival, same de jure empire, nearby, or in the war) are notified — the trigger's evident intent. Required overriding `09_mpo_wars.txt` (added unmodified first).

**Removed the `ck3-tiger.conf` scope ignore** — no longer needed.

**Shortened the `#Unop` comment** on the imprisonment guard.

One new, unrelated tiger error surfaced from now overriding `09_mpo_wars.txt`: `attacker_wargoal_percentage = 1.5` (>1) on a Domination-war CB at line 3529. That's a pre-existing vanilla data quirk; clamping it would change war pacing, so I left it precisely `# Ignored` and flagged it for its own issue rather than rebalance here.

`make tiger`: clean.

ck3-tiger.conf
pharaox: **fixer:** Per maintainer request, this scope warning is now *fixed* rather than ignored — parameterized the trigger with `$GOK$` and pass `scope:defender` from the `09_mpo_wars.txt` on_victory, so `root` (the casus belli) is no longer used as a character. This ignore rule has been removed.

pharaox:
`@fixer` Commit these 2 to memory:

* **Never** ignore real issues reported by `tiger`. **Always** fix them. You can ignore only if they don't actually have impact.
* **Never** change vanilla objects interfaces. If vanilla adds a new usage in a future version, this would silently break Unop. Consider alternative fixes, or add a `unop_` counterpart of the vanilla object.

Now correct the above fix.

pharaox:
**fixer:** Corrected per the two memory rules.

**Don't change the vanilla object's interface.** Reverted `mpo_war_of_defiance_notified_player_trigger` to byte-identical vanilla and instead added a `unop_war_of_defiance_notified_player_trigger` counterpart (same body, `root` → `$GOK$` parameter). The `09_mpo_wars.txt` `on_victory` loop now calls the `unop_` trigger with `GOK = scope:defender`, so the connection checks run against the new Gurkhan, not the casus belli — fixing the scope bug without mutating a vanilla object (a future vanilla call to the original trigger keeps working).

**Tiger.** The scope warning stays fixed (the `unop_` trigger uses `$GOK$`, not `root`). The remaining `attacker_wargoal_percentage = 1.5` warning is **intended vanilla design, not a bug**: the same `1.5` is used by vanilla's nomadic-conquest CBs (`00_nomadic_conquest.txt`) — it makes the attacker never tick warscore from occupation, so these conquest wars are decided by battles/total occupation. Clamping it to `≤ 1` would change intended war pacing, so it stays precisely `# Ignored` with a corrected justification. `make tiger`: clean.

pharaox:
**fixer:** Both rules committed to memory and applied:

1. *Never ignore real tiger bugs — fix them; ignore only when there's no real impact.* The scope bug is fixed (not ignored). The one remaining ignore (`attacker_wargoal_percentage = 1.5`) is intended vanilla design replicated from the nomadic-conquest CBs, not a bug — so it's correctly ignored.
2. *Never change a vanilla object's interface.* Done — the vanilla trigger is restored byte-identical and a `unop_` counterpart with the `$GOK$` parameter is called from the overridden call site instead.

ck3-tiger.conf
pharaox: **fixer:** Update: the scope bug is now fixed via a `unop_war_of_defiance_notified_player_trigger` counterpart (vanilla trigger restored byte-identical), not an ignore. The only ignore left in this file is the `attacker_wargoal_percentage = 1.5` `range` warning, which is intended vanilla design (same value used by the nomadic-conquest CBs) rather than a bug — clamping it would change war pacing.

common/scripted_triggers/09_mpo_greatest_of_khans_triggers.txt
```diff
@@ -197,7 +197,50 @@ mpo_gok_submitting_coward_trigger = {
...
+			exists = root.capital_county
+			realm_to_title_distance_squared = {
+				target = root.capital_county
+				value <= squared_distance_almost_massive
+			}
+		}
+	}
+}
+
+unop_war_of_defiance_notified_player_trigger = {
```
pharaox: `@fixer` Move this trigger to its proper `unop_` file.

pharaox: **fixer:** Done — moved `unop_war_of_defiance_notified_player_trigger` to `common/scripted_triggers/unop_triggers.txt`. The `09_mpo_greatest_of_khans_triggers.txt` override now differs from vanilla only by the one-line #559 imprisonment guard.

pharaox:
**fixer:** Moved the `unop_` trigger to its proper Unop file.

- `unop_war_of_defiance_notified_player_trigger` now lives in `common/scripted_triggers/unop_triggers.txt` (the general Unop trigger home), not appended to the `09_mpo_greatest_of_khans_triggers.txt` vanilla override.
- That override is now byte-identical to vanilla apart from the single #559 imprisonment guard line.
- Call site in `09_mpo_wars.txt` unchanged (`unop_war_of_defiance_notified_player_trigger = { GOK = scope:defender }`).

`make tiger`: clean.

PR #574, opened 2026-06-28
Fix reversed Dynastic Cycle expansion catalyst polarity during Expansion phase

### LINKED ISSUE #551: Dynastic Cycle 'push towards/away from Expansion' catalysts apply to wrong era during Expansion
**Reported on Steam by `Nider001` on `2026-06-26 14:26`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> The "House Head consulted Heaven — push towards Expansion" and "Movement consulted Heaven — push towards Expansion" Dynastic Cycle catalysts contribute towards Tension era while the current era is Expansion. Meanwhile, both respective "... — push away from Expansion" catalysts contribute away from Tension era. This setup just doesn't make sense - since the current era is Expansion, the catalyst effects should be reversed. I didn't check if the same happens during Advancement era, but it's worth investigating as well.

pharaox:
Fixes #551

**fixer:** In the **stability_expansion → instability** `future_phases` block of `dynastic_cycle`, the four "consulted Heaven" *expansion* catalysts had reversed polarity:

- `..._expansion_positive` ("push **towards** Expansion") was set to `_gain` — i.e. it accelerated decay *out of* Expansion toward Instability.
- `..._expansion_negative` ("push **away from** Expansion") was set to `_loss` — i.e. it resisted that decay.

Per the catalyst loc and the meaning of `gain` (progress toward the target phase = Instability) vs `loss` (resist), this is backwards while the current era is Expansion. Flipped the direction keyword on those four lines (keeping each line's magnitude tier), so `_expansion_positive` now resists the transition (`_loss`) and `_expansion_negative` accelerates it (`_gain`). This makes them mirror the correctly-polarized `_instability_*` siblings in the same block.

The identical-looking blocks in the *instability → stability_expansion* transitions (where `_expansion_positive = gain` correctly accelerates entering Expansion) are **unchanged** — edited by line position. Advancement era doesn't use these catalysts in its decay block, so it's unaffected (answers the reporter's question).

**Tiger:** adding the situation file surfaced two pre-existing, unrelated vanilla scope warnings (`trigger_situation_catalyst { character = root }` in two phase `on_start` blocks, where `root` is the situation per `_situations.info`). Separate vanilla bug; ignored via a precise `ck3-tiger.conf` rule and flagged for follow-up. `make tiger` clean.

### DIFF
--- a/ck3-tiger.conf
+++ b/ck3-tiger.conf
@@ -597,6 +597,11 @@ filter = {
 			text = "expects root to be character but root seems to be faith"
 			file = common/scripted_triggers/07_frankokratia_triggers.txt
 		}
+		NAND = { # Pre-existing vanilla bug: phase on_start root is the situation (per _situations.info), but trigger_situation_catalyst's `character = root` passes the situation where a character is expected; a separate fix, out of scope here
+			key = scopes
+			text = "`root` produces situation but expected character"
+			file = common/situation/situations/tgp_dynastic_cycle.txt
+		}
 		NAND = { # Ignore scope warnings in event backgrounds, too difficult to fix
 			key = scopes
 			OR = {
--- a/ck3-tiger.conf
+++ b/ck3-tiger.conf
@@ -597,11 +597,6 @@ filter = {
 			text = "expects root to be character but root seems to be faith"
 			file = common/scripted_triggers/07_frankokratia_triggers.txt
 		}
-		NAND = { # Pre-existing vanilla bug: phase on_start root is the situation (per _situations.info), but trigger_situation_catalyst's `character = root` passes the situation where a character is expected; a separate fix, out of scope here
-			key = scopes
-			text = "`root` produces situation but expected character"
-			file = common/situation/situations/tgp_dynastic_cycle.txt
-		}
 		NAND = { # Ignore scope warnings in event backgrounds, too difficult to fix
 			key = scopes
 			OR = {
--- a/common/situation/situations/tgp_dynastic_cycle.txt
+++ b/common/situation/situations/tgp_dynastic_cycle.txt
@@ -1,0 +1,1965 @@
+﻿@gained_mandate_of_heaven = 4000
+@gained_mandate_of_heaven_halved = 2000
+@lost_mandate_of_heaven = 4000
+@lost_mandate_of_heaven_halved = 2000
+@mandate_of_heaven_flavorful_points = @[gained_mandate_of_heaven + 1000]
+
+dynastic_cycle = {
+	window = dynastic_cycle
+	gui_window_name = "window_tgp_dynastic_cycle"
+	gui_tooltip_group_focused = yes
+	map_mode = dynastic_cycle
+	illustration = "gfx/interface/illustrations/event_story/tgp_dynastic_cycle_corruption.dds"
+	situation_group_type = major
+
+	is_unique = yes
+	keep_full_history = yes 
+	
+	##################################################
+	# Regions
+	##################################################
+	sub_regions = {
+		core = {
+			icon = "gfx/interface/icons/situation_types/dynastic_cycle.dds"
+		}
+	}
+
+	##################################################
+	# On Actions
+	##################################################
+	on_monthly = {
+		trigger_event = {
+			on_action = dynastic_cycle_monthly_pulse
+		}
+	}
+	on_yearly = {
+		trigger_event = {
+			on_action = dynastic_cycle_yearly_pulse
+		}
+		#update all movement leaders
+		every_participant_group = {
+			save_scope_as = situation_participant_group
+			if = {
+				limit = { # Appoint a new movement leader if no longer alive
+					exists = var:movement_leader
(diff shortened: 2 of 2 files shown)


pharaox:
`@fixer` Fix the pre-existing vanilla bug.

common/situation/situations/tgp_dynastic_cycle.txt
```diff
@@ -0,0 +1,1965 @@
...
+						### SHARED CORRUPTION CATALYSTS
+
+						# Movement Power
+						catalyst_conservative_movement_yearly = massive_situation_catalyst_over_time_gain
+						catalyst_expansion_movement_yearly = major_situation_catalyst_loss
+						catalyst_movement_gained_power_advancement = major_situation_catalyst_gain
+						catalyst_movement_gained_power_expansion = medium_situation_catalyst_loss
+						catalyst_movement_gained_power_conservative = major_situation_catalyst_gain
+						catalyst_movement_gained_power_pro_hegemon = minimal_situation_catalyst_loss
+						catalyst_movement_consulted_heaven_expansion_positive = massive_situation_catalyst_loss #Unop reversed polarity: during Expansion, pushing TOWARDS Expansion should resist decay to Instability, not accelerate it
```
pharaox, line 715: **reviewer:** Correct polarity flip and correctly localized. This is the `stability_expansion → instability` decay block (phase def at line 628), the lone outlier; the byte-identical instability→expansion blocks (1298-1301, 1606-1609) keep the original `positive=gain` and are rightly untouched. `_loss = _gain × -1` in `10_tgp_dynastic_cycle_values.txt`, so magnitude is preserved and only the sign changes.

ck3-tiger.conf
```diff
@@ -597,6 +597,11 @@ filter = {
 			text = "expects root to be character but root seems to be faith"
 			file = common/scripted_triggers/07_frankokratia_triggers.txt
 		}
+		NAND = { # Pre-existing vanilla bug: phase on_start root is the situation (per _situations.info), but trigger_situation_catalyst's `character = root` passes the situation where a character is expected; a separate fix, out of scope here
```
pharaox: **reviewer:** Verified genuine: `trigger_situation_catalyst { character = root }` at lines 1214/1531 runs in phase `on_start` where `_situations.info` says `Root = situation`, so this is a real pre-existing vanilla scope bug unrelated to the polarity fix. `# Ignored` is correct (tiger isn't wrong); deferring to its own issue is the right call.

pharaox (commented):
**reviewer:** Looks good — flips `gain`↔`loss` on the four `expansion_*` consult-Heaven catalysts in the `stability_expansion → instability` decay block (lines 715/716/719/720), keeping each magnitude tier. Confirmed in `10_tgp_dynastic_cycle_values.txt` that `_loss = _gain × -1`, so it's a pure sign flip: "towards Expansion" now resists decay and "away" accelerates it, matching the loc. `diff` vs vanilla shows exactly those four lines changed — the byte-identical instability→expansion sibling blocks (1298-1301, 1606-1609) are correctly untouched. Advancement era genuinely lacks the bug (uses `instability_*` catalysts). Vanilla file added byte-identical in its own commit.

Independently verified the tiger ignore: the two `root produces situation but expected character` warnings are `trigger_situation_catalyst { character = root }` in phase `on_start` blocks where `_situations.info` confirms `Root = situation` — a real pre-existing vanilla scope bug, unrelated and out of scope; correct under `# Ignored`, well grouped.

- Low: that deferred scope bug deserves its own follow-up issue, as flagged — not blocking.

pharaox:
**fixer:** Fixed the pre-existing vanilla scope bug instead of ignoring it (`@fixer` request).

The two warnings were `trigger_situation_catalyst = { catalyst = catalyst_hegemon_lost_mandate_of_heaven; character = root }` in the `instability_conquest` (line ~1214) and `instability` (line ~1531) phase `on_start` blocks. Per `_situations.info`, a phase `on_start` has `Root = situation`, so `character = root` passed the **situation** where a **character** is expected — and these blocks only run when `NOT exists title:h_china.holder`, i.e. there's no current hegemon character to attribute anyway.

`character` is the optional history-attribution argument (the epidemic catalysts a few lines up in this same file use the short `trigger_situation_catalyst = catalyst_X` form with no character). Fix: comment out the invalid `character = root` so the catalyst still fires/records identically, just without the bogus attribution. No behavior change beyond removing the invalid scope.

Removed the now-unneeded `ck3-tiger.conf` scope ignore. `make tiger`: clean.

The expansion-polarity fix is unchanged.

ck3-tiger.conf
pharaox: **fixer:** Per maintainer request, this is now *fixed* rather than ignored — the invalid `character = root` arg is dropped (phase on_start root is the situation, not a character; attribution is optional, matching the epidemic catalysts in the same file). This ignore rule has been removed.

ProZeratul (approved):

Issue #479, opened 2026-05-23
friend_alliance / friend_alliance_corresponding loc keys swap the two characters


pharaox:
**Reported on Steam by `Nider001` on `2026-05-19 12:29`** ([link](https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329)).

> The "friend_alliance" and "friend_alliance_corresponding" locs (which are used in "allies_progress_towards_friendship_effect") currently mention the two characters backwards. Let's say Bob joined Sam's war as a secondary attacker/defender via alliance or Join War interaction. When the war ends and they become friends, the friendship reason will say "Sam heeded many of Bob's calls to war, proving that a valuable alliance had been formed." This is incorrect.

pharaox:
## Investigation

**Files checked:**
- `localization/english/relationship_reasons_l_english.yml:107-108` (vanilla) — `friend_alliance` / `friend_alliance_corresponding`.
- `common/scripted_effects/00_war_effects.txt` — `allies_progress_towards_friendship_effect` calls `progress_towards_friend_effect` with `CHARACTER = scope:ally` inside `scope:actor` (war leader).

**Findings:**
- The relationship is set on the war leader with target = ally, so in the loc `CHARACTER` = war leader and `TARGET_CHARACTER` = ally.
- Current text "[CHARACTER] heeded many of [TARGET_CHARACTER]'s calls to war" therefore reads "leader heeded ally's calls", which inverts the actual event.
- Fix is loc-only: swap `CHARACTER` and `TARGET_CHARACTER` in both keys so the ally is the subject. No script change needed.

**Classification:** `confirmed` / `vanilla` / `localization`

## Fix in Progress

A fix is being prepared. PR will follow shortly.

PR #582, opened 2026-07-12
Fix reversed friend_saved_as_children and rival_lost_duel relationship reasons

### LINKED ISSUE #575: friend_saved_as_children / rival_lost_duel relationship reason keys appear reversed with their _corresponding variants
**Reported on Steam by `吉吉` on `2026-07-01 11:52`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> Sorry to bother you, but I'm not sure if this is a problem or if it's just that my English isn't good enough—I feel like there's something off about the way the characters' relationships are described. (I don't usually use the English version, but when I switched to it, I noticed the same issue.)
>
>  https://imgur.com/a/wtvNKf5
>
> Specifically, I think the [friend_saved_as_children_corresponding] and [friend_saved_as_children] in the [localization/english/relationship_reasons_l_english.yml] file seem to be reversed.
>
> [rival_lost_duel] and [rival_lost_duel_corresponding] seem to have the same issue, but I haven't encountered it in the game recently, so I don't have any screenshots.

Screenshot: https://i.imgur.com/N0JwY8d.jpeg

pharaox:
Fixes #575

**fixer:** The base/`_corresponding` strings are swapped for two relationship reasons. By vanilla convention the base key credits `target` (→ TARGET_CHARACTER) as the doer and the `_corresponding` key credits the running character (→ CHARACTER).

- `friend_saved_as_children` is set by the saver with `target =` the saved child, so the base should credit TARGET_CHARACTER as the saver — vanilla credited CHARACTER.
- `rival_lost_duel` is set in `single_combat.0031.c` by the loser with `target =` the victor, so the base should read "victor bested loser" — vanilla read "loser bested victor".

Fixed by swapping the base/corresponding strings for each pair in `unop_fixed_keys_l_english.yml`, preserving the vanilla version numbers. Mirrors the earlier #479 fix for `friend_alliance`.


ProZeratul:
There are a lot of those so it won't be the last time we see this kind of errors

PR #543, opened 2026-06-20
Remove Siberian permafrost modifier when conquering the North via Expand the Steppe

pharaox:
Fixes #534

**fixer:** The `mpo_siberian_permafrost_modifier(_bad)` loc states the penalty is removed "after conquering the whole North by the expand_the_steppe_decision for the dlc_mpo_steppe_siberia_further_expansion region", but the Siberia branch of `expanding_steppe_effect` only added the region to the steppe situation and never removed the modifier.

Added the removal in that branch: iterate `every_county_in_region = { region = mpo_region_permafrost }` (the same region the modifier is added to at game start) and remove whichever variant (`mpo_siberian_permafrost_modifier` or `_bad`) is present, mirroring the existing removal idiom in vanilla `00_decisions_effects.txt`.

In-game verification needed: complete the Siberia expansion (`completely_controls_region = dlc_mpo_steppe_siberia_further_expansion`) and confirm the permafrost modifiers are removed from the permafrost counties.

PR #584, opened 2026-07-12
Clear Siberian permafrost debuff when Aid Tribal Development feudalizes a holding

### LINKED ISSUE #577: Siberian permafrost debuff not removed by Aid Tribal Development decision
**Reported on Steam by `Sledjer` on `2026-07-01 02:19`** (https://steamcommunity.com/workshop/filedetails/discussion/2871648329/4697908845434539387/).

> Is there any chance you fix Siberian permafrost sticking around after the aid tribal development decision?
> https://forum.paradoxplaza.com/forum/threads/the-siberian-permafrost-debuff-cannot-be-removed-by-aid-tribal-development.1889349/

pharaox:
Fixes #577

**fixer:** "Aid Tribal Development" (`support_feudalize_tribal_holding_interaction`) feudalizes a vassal's tribal holding but never removed the Siberian permafrost county modifier, so the `-6 development_decline` debuff persisted.

The sibling `feudalize_holding_interaction` (feudalizing your *own* holding) already clears it: on a permafrost-region county it removes `mpo_siberian_permafrost_modifier`/`_bad`. `support_feudalize` performs the same holding conversion but omitted this cleanup — the lone outlier among the tribal→feudal conversion paths.

Fix: mirror the sibling's permafrost-removal block in `support_feudalize`'s `on_accept`, after the holding is converted (wrapped in `hidden_effect`, scoped on `scope:target`). Handles both the castle and mandala temple_citadel branches (added once after both).

Adding the file as a new override surfaced one latent vanilla strict-scopes warning (`feudalize_holding_interaction_cost` in `ai_potential`); it's a false positive — the cost value reads `scope:target.county` with the safe `?=` operator — so it's added to `ck3-tiger.conf` under False positives. `make tiger` clean.

Vanilla file added unmodified in a separate commit.

### DIFF
--- a/ck3-tiger.conf
+++ b/ck3-tiger.conf
@@ -109,6 +109,11 @@ filter = {
 			text = "`diarch_overwrites_old_designated_heir_apply_opinions_effect` expects scope:ai_recipient to be set"
 			file = common/character_interactions/06_ep3_interactions.txt
 		}
+		NAND = { # feudalize_holding_interaction_cost reads scope:target.county with ?=, so it's safe when scope:target isn't set (e.g. ai_potential)
+			key = strict-scopes
+			text = "`feudalize_holding_interaction_cost` expects scope:target to be set"
+			file = common/character_interactions/00_tribal_interactions.txt
+		}
 		NAND = { # scheme scopes are generated automatically
 			key = strict-scopes
 			OR = {
--- a/ck3-tiger.conf
+++ b/ck3-tiger.conf
@@ -109,11 +109,6 @@ filter = {
 			text = "`diarch_overwrites_old_designated_heir_apply_opinions_effect` expects scope:ai_recipient to be set"
 			file = common/character_interactions/06_ep3_interactions.txt
 		}
-		NAND = { # feudalize_holding_interaction_cost reads scope:target.county with ?=, so it's safe when scope:target isn't set (e.g. ai_potential)
-			key = strict-scopes
-			text = "`feudalize_holding_interaction_cost` expects scope:target to be set"
-			file = common/character_interactions/00_tribal_interactions.txt
-		}
 		NAND = { # scheme scopes are generated automatically
 			key = strict-scopes
 			OR = {
--- a/common/character_interactions/00_tribal_interactions.txt
+++ b/common/character_interactions/00_tribal_interactions.txt
@@ -692,6 +692,26 @@ support_feudalize_tribal_holding_interaction = {
 				}	
 			}
 		}
+		#Unop Feudalizing the holding through this interaction must also clear the Siberian permafrost debuff, mirroring feudalize_holding_interaction
+		hidden_effect = {
+			scope:target ?= {
+				if = {
+					limit = {
+						title_province = {
+							geographical_region = geographical_region:mpo_region_permafrost
+						}
+					}
+					if = {
+						limit = { has_county_modifier = mpo_siberian_permafrost_modifier }
+						remove_county_modifier = mpo_siberian_permafrost_modifier
+					}
+					else_if = {
+						limit = { has_county_modifier = mpo_siberian_permafrost_modifier_bad }
+						remove_county_modifier = mpo_siberian_permafrost_modifier_bad
+					}
+				}
+			}
+		}
 		scope:actor = {
 			# If you are admin, you get some influence for your efforts
 			if = {
--- a/common/character_interactions/00_tribal_interactions.txt
+++ b/common/character_interactions/00_tribal_interactions.txt
@@ -741,7 +741,9 @@ support_feudalize_tribal_holding_interaction = {
 		is_at_war = no
 		is_adult = yes
 		# The AI only does this if they have plenty of gold to spare
-		gold >= { value = feudalize_holding_interaction_cost multiply = 4 }
+		#Unop feudalize_holding_interaction_cost reads scope:target.county, which is not set here in ai_potential, so its permafrost surcharge branch never applied; use the resolved base cost (500) x4 directly
+		#gold >= { value = feudalize_holding_interaction_cost multiply = 4 }
+		gold >= 2000
 		# Warring AI don't do this - They want to spend their gold on war!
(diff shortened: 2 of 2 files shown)


pharaox:
`@fixer` Fix the tiger issue rather than dismissing it as "safe", it looks like a real bug. `scope:target` is not available in `ai_potential`, so whatever the author intended is broken. Either replace with a condition that can be evaluated (e.g. using a hardcoded gold value), or remove it entirely.

common/character_interactions/00_tribal_interactions.txt
```diff
@@ -0,0 +1,952 @@
...
+								}
+							}
+						}
+					}	
+				}	
+			}
+		}
+		#Unop Feudalizing the holding through this interaction must also clear the Siberian permafrost debuff, mirroring feudalize_holding_interaction
+		hidden_effect = {
+			scope:target ?= {
```
pharaox, line 697: **reviewer:** Correct — exact structural mirror of the `feudalize_holding_interaction` template (lines 185–201), placed after both the mandala temple_citadel and castle conversion branches, inside the outer `scope:recipient`. The `scope:target ?=` and `title_province` region check match the template; the real `set_holding_type` has already run in the visible toasts, so the county is feudal here. `hidden_effect` keeps it out of the accept toast, matching PR #543's approach. Good.

ck3-tiger.conf
```diff
@@ -109,6 +109,11 @@ filter = {
 			text = "`diarch_overwrites_old_designated_heir_apply_opinions_effect` expects scope:ai_recipient to be set"
 			file = common/character_interactions/06_ep3_interactions.txt
 		}
+		NAND = { # feudalize_holding_interaction_cost reads scope:target.county with ?=, so it's safe when scope:target isn't set (e.g. ai_potential)
```
pharaox: **reviewer:** Genuine false positive, verified: removing this rule re-surfaces exactly one `strict-scopes` warning — `feudalize_holding_interaction_cost` used in `ai_potential` (line 744), whose definition reads `scope:target.county ?=`, which is safe when `scope:target` isn't set. Surfaced only because the file is now overridden. Correctly placed under `# False positives` next to the sibling strict-scopes rule, with an accurate comment. No paperover.

pharaox (commented):
**reviewer:** Looks good — the fix mirrors the exact permafrost-removal block from the sibling `feudalize_holding_interaction` into `support_feudalize_tribal_holding_interaction.on_accept`, wrapped in `hidden_effect`. Verified placement: it runs after both real `set_holding_type` branches (in the visible toasts; the nested `show_as_tooltip` copies are display-only), inside the outer `scope:recipient`, with explicit `scope:target ?=` — so the holding is feudal by the time it clears the modifier. Pattern-class audit confirms `support_feudalize` was the lone tribal→feudal path missing the cleanup. Vanilla file added byte-identical in its own commit; #Unop comment correct.

The added `ck3-tiger.conf` false-positive is genuine: I removed it and tiger re-surfaced exactly one strict-scopes warning (`feudalize_holding_interaction_cost` in `ai_potential`), whose script-value reads `scope:target.county ?=` — safe under `?=`. Correctly classified and placed. `make tiger`: no problems found.

- Low: adjacent `feudalize_nomadic_holding_interaction` / `feudalize_herder_holding_interaction` (09_mpo_interactions.txt) omit the same cleanup for nomad/herder holdings — out of scope for #577, worth a separate issue.
- Low: verified by reading the code but untested in-game; recommend the confirmation the Fix section notes.

pharaox:
**fixer:** Agreed — it was a real bug, not a false positive. In `ai_potential`, `scope:target` isn't set, so the `scope:target.county` permafrost-surcharge branch inside `feudalize_holding_interaction_cost` could never contribute, silently breaking the intended gold gate. Removed the false-positive rule from `ck3-tiger.conf` and replaced the `ai_potential` check with the resolved hardcoded value `gold >= 2000` (the script value's base `500` × 4, matching what it evaluates to when `scope:target` is absent). `make tiger` is clean.

Also rebased the branch onto `main` (it had fallen behind).

common/character_interactions/00_tribal_interactions.txt
```diff
@@ -0,0 +1,954 @@
...
+	}
+	
+	ai_potential = {
+		# Some basic triggers
+		is_at_war = no
+		is_adult = yes
+		# The AI only does this if they have plenty of gold to spare
+		#Unop feudalize_holding_interaction_cost reads scope:target.county, which is not set here in ai_potential, so its permafrost surcharge branch never applied; use the resolved base cost (500) x4 directly
+		#gold >= { value = feudalize_holding_interaction_cost multiply = 4 }
+		gold >= 2000
```
pharaox, line 746: **reviewer:** Confirmed correct. `_character_interactions.info` (l.545–547) documents that in `ai_potential` *"Root scope is the actor, and other event targets are not available"* — so `feudalize_holding_interaction_cost`'s `scope:target.county ?=` permafrost branch (`+2000`) could never fire here, and the base `500 × 4 = 2000` is the faithful resolved value. One nuance (Low, not blocking): the script value also has an `is_ai = yes` + tribal-capital `add = -250` branch, but this interaction's actor is a non-tribal/non-nomadic liege whose `capital_province` is a castle, so `-250` effectively never applies — 2000 is right.

pharaox (commented):
**reviewer:** Looks good — Fix 2 correctly resolves the `ai_potential` scope bug that my Review 1 wrongly endorsed as a false positive. `_character_interactions.info` confirms `scope:target` is not available in `ai_potential`, so the `+2000` permafrost surcharge in `feudalize_holding_interaction_cost` never applied there; the gate was always `500 × 4 = 2000`. Removing the suppression and hardcoding `gold >= 2000` fixes it properly. Diff vs vanilla is minimal (only the two `#Unop` changes), commit hygiene is clean, tiger reports no problems.

- Low: hardcoded 2000 ignores the script value's `is_ai` tribal-capital `-250` branch, but that never applies to this non-tribal-liege actor — matches vanilla for all realistic cases.

PR #585, opened 2026-07-12
Remove disqualified tournament saboteur from the bracket

### LINKED ISSUE #520: Grand Tournament: disqualified saboteurs keep competing; event texts imply wrong outcomes
**Reported on Steam by `The Emperor of His Room` on `2026-05-27 06:36`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> Are you guys thinking of doing a pass on grand tournaments? The events during the tournament constantly imply things happen when they don't like being disqualified or increasing your chances of winning when it actually drops.

pharaox:
Fixes #520 (Bug 1 — saboteur keeps competing)

**fixer:** A saboteur caught in `tournament_events.6051` while a **qualified** contestant only got the disqualified log + `disqualified_contest_flag` (2-month) + aptitude removal — he was never removed from the bracket, so he kept being matched and could reach the finals (confirmed by the reporter's screenshots).

The disqualification relied entirely on the opponent-side `disqualified_contest_flag` bye in `contest_events.0820`, but that flag lasts only `months = 2` and isn't renewed, while a Grand Tournament plays out over several 1-month versus rounds — so by later rounds the flag has lapsed and the saboteur competes normally.

Fix: in the qualified-saboteur branch, drop his `progress_to_victory` and `remove_from_activity` — mirroring what the sibling (spectator-saboteur) `else` branch already does for a caught saboteur. Removing him from the activity makes the next round's matchmaking exclude him, so he provably can't reach later rounds. Keeps the existing disqualified log/flag/opinion for flavor. `make tiger` clean.

**Scope (deliberately limited):** This PR fixes only Bug 1 (the titled bug). Bugs 2 (spectator text implies active participation) and 3 (skill-check option text vs. confirm tooltip) from the original report are real but not yet located to specific events/options — they should be split into a separate issue with the exact event IDs / option texts.

### DIFF
--- a/events/activities/tournaments/tournament_events.txt
+++ b/events/activities/tournaments/tournament_events.txt
@@ -14635,9 +14635,8 @@ tournament_events.6051 = { # Sabotage Outcome
 						}
 					}
 				}
-				#Unop Only the aptitude was removed, which excludes the saboteur from turn-based contests but not from
-				#Unop versus brackets, where pairing ignores it, so he kept being matched and could reach the finals.
-				#Unop Route by phase to the proper disqualification effect, as the sibling event tournament_events.6052 does.
+				#Unop Only the aptitude was removed, so the saboteur kept being matched and could reach the finals.
+				# Route by phase to the proper disqualification effect, as the sibling event tournament_events.6052 does.
 				#remove_variable = contest_aptitude
 				#add_character_flag = {
 				#	flag = disqualified_contest_flag #Unop: Use correct flag
--- a/events/activities/tournaments/tournament_events.txt
+++ b/events/activities/tournaments/tournament_events.txt
@@ -14635,33 +14635,58 @@ tournament_events.6051 = { # Sabotage Outcome
 						}
 					}
 				}
-				remove_variable = contest_aptitude
-				add_character_flag = {
-					flag = disqualified_contest_flag #Unop: Use correct flag
-					months = 2
-				}
+				#Unop Only the aptitude was removed, which excludes the saboteur from turn-based contests but not from
+				#Unop versus brackets, where pairing ignores it, so he kept being matched and could reach the finals.
+				#Unop Route by phase to the proper disqualification effect, as the sibling event tournament_events.6052 does.
+				#remove_variable = contest_aptitude
+				#add_character_flag = {
+				#	flag = disqualified_contest_flag #Unop: Use correct flag
+				#	months = 2
+				#}
+				#scope:activity = {
+				#	add_activity_log_entry = {
+				#		key = tournament_disqualified_log
+				#		score = 50
+				#		tags = { contestant removal }
+				#		character = scope:sabotage_actor
+				#		target = scope:host
+				#
+				#		# Effects
+				#		custom_description = {
+				#			text =  tournament_contest_disqualify_tt
+				#			subject = scope:sabotage_actor # DISQUALIFIED
+				#		}
+				#		scope:sabotage_actor = {
+				#			if = {
+				#				limit = { is_ai = yes }
+				#				add_opinion = {
+				#					target = scope:host
+				#					modifier = tournament_disqualified_opinion
+				#					opinion = -15
+				#				}
+				#			}
+				#		}
+				#	}
+				#}
 				scope:activity = {
-					add_activity_log_entry = {
-						key = tournament_disqualified_log
-						score = 50
-						tags = { contestant removal }
-						character = scope:sabotage_actor
-						target = scope:host
-
-						# Effects
-						custom_description = {
-							text =  tournament_contest_disqualify_tt
-							subject = scope:sabotage_actor # DISQUALIFIED
+					switch = {
+						trigger = has_current_phase
+						# Versus Contests require more specific script
+						tournament_phase_joust = {
+							tournament_contest_versus_disqualification_effect = { DISQUALIFIED = scope:sabotage_actor MATCH = scope:sabotage_target CONTEST = joust SKILL = horse }
 						}
-						scope:sabotage_actor = {
-							if = {
-								limit = { is_ai = yes }
-								add_opinion = {
-									target = scope:host
-									modifier = tournament_disqualified_opinion
-									opinion = -15
-								}
-							}
+						tournament_phase_wrestling = {
+							tournament_contest_versus_disqualification_effect = { DISQUALIFIED = scope:sabotage_actor MATCH = scope:sabotage_target CONTEST = wrestling SKILL = pugilism }
+						}
+						tournament_phase_duel = {
+							tournament_contest_versus_disqualification_effect = { DISQUALIFIED = scope:sabotage_actor MATCH = scope:sabotage_target CONTEST = duel SKILL = foot }
+						}
+						tournament_phase_board_game = {
+							tournament_contest_versus_disqualification_effect = { DISQUALIFIED = scope:sabotage_actor MATCH = scope:sabotage_target CONTEST = board_game SKILL = board_game }
+						}
+						# Non-Versus contests
+						fallback = {
+							tournament_contest_disqualification_effect = { DISQUALIFIED = scope:sabotage_actor }
 						}
 					}
 				}


events/activities/tournaments/tournament_events.txt
```diff
@@ -14665,6 +14666,10 @@ tournament_events.6051 = { # Sabotage Outcome
 						}
 					}
 				}
+				#Unop A caught saboteur was only flagged, never removed from the bracket, so he kept being matched and could reach the finals; eject him from the activity like a caught spectator saboteur, so later-round matchmaking excludes him
+				scope:sabotage_actor = {
+					remove_from_activity = scope:activity
```
pharaox: **reviewer:** Correct and the key line. Verified `scope:sabotage_actor` is the saboteur and is `root` of this chain, and that ejecting him excludes him from later-round matchmaking. His qualified-round opponent still advances via the retained `disqualified_contest_flag` in `contest_events.0820` (its `versus_contestant` resolves from a stored match variable, which survives `remove_from_activity`). Mirrors the spectator-saboteur `else` branch's proven use of `remove_from_activity`.

pharaox (commented):
**reviewer:** Looks good — the fix ejects the caught qualified saboteur from the activity (`remove_from_activity`) and strips his `progress_to_victory`, so next-phase matchmaking (which draws from remaining subset members) excludes him and he can no longer reach the finals. Opponent advancement is left to the existing vanilla mechanism: the PR keeps `disqualified_contest_flag`, and the opponent's `contest_events.0820` resolves `versus_contestant` from a stored match variable (survives removal) and takes the 'Opponent disqualified' bye. Declining to re-derive `$MATCH$`/`$CONTEST$`/`$SKILL$` to call the versus-disqualification effect in 6051 is the right, lower-risk call. `remove_from_activity` is already the proven removal in this event's spectator-saboteur branch. Scope is Bug 1 only, as stated. Tiger clean; #Unop comments correct.

- Low: relies on the saboteur's qualified-round opponent hitting 0820 within the flag's 2-month life to take the bye — holds under normal round cadence and is owned by the pre-existing vanilla flag path, not this fix. Recommend the in-game check the Fix section notes. Not a blocker.

events/activities/tournaments/tournament_events.txt
ProZeratul: This will prevent the contestant to participate in all other event in this tournament, so if he cheat on the first one he will be prevented to participate in all the other contests wich is not what we want here

ProZeratul (changes requested):
This will prevent the contestant to participate in all other event in this tournament, so if he cheat on the first one he will be prevented to participate in all the other contests wich is not what we want here.

If we take this route, we need a way to re-invite this contestant right after. I've tried to make a clean version where the character is still in the contest but marked to fail but it always break in the following events.

pharaox:
`@analyzer` Re-analyze taking the feedback above into account.

pharaox:
**fixer:** @ProZeratul You were right — the `remove_from_activity` approach is retracted and the branch force-pushed with a different fix.

**Why your objection was correct.** A tournament is one activity with many contest phases, and each phase re-composes its own `qualified` set from its `contestant` subset at `on_phase_active`. So `remove_from_activity` evicts the saboteur from the *whole tournament*, barring him from every later contest — exactly what you described. Vanilla itself corroborates this: the sibling event `tournament_events.6052` reserves `remove_from_activity` for its `else` branch only, i.e. for a saboteur who is *not* a contestant.

**The actual fix: restore what vanilla already wrote next door.** `6050` dispatches to `6052` when the host is a human player, and to `6051` otherwise — and `6051` is a degraded copy of `6052` with the phase switch dropped. `6051` only did `remove_variable = contest_aptitude`, which works for turn-based contests (winner selection filters on `exists = var:contest_aptitude`) but **not** for versus brackets: `tournament_contest_versus_pair_effect` iterates the subset with no aptitude and no flag check, and advancement is additive, so a disqualified saboteur still in `qualified` keeps getting paired and can reach the finals. That's the reporter's screenshots.

`6051`'s qualified branch now uses the same `switch = { trigger = has_current_phase … }` block as `6052`: versus phases call `tournament_contest_versus_disqualification_effect`, non-versus falls back to `tournament_contest_disqualification_effect`.

This also explains the *"script broke and nobody was selected as winner"* you hit while experimenting: removing the saboteur without advancing anyone leaves the stage underpopulated, so `tournament_contest_versus_winner_effect` finds no `winner` subset member. The versus effect avoids that by `add_to_guest_subset`-ing the opponent into the next stage in his place.

I dropped `6051`'s inlined log entry, because `tournament_contest_versus_disqualification_effect` writes its own `tournament_disqualified_log` — keeping both would double-log. Both effects also set `disqualified_contest_flag` and strip aptitude + `progress_to_victory` themselves, so nothing is lost.

**One scope question for you.** I pass `MATCH = scope:sabotage_target`, mirroring `6052` exactly. That assumes the sabotage victim is the saboteur's bracket opponent, which isn't guaranteed — `6050` saves up to 3 targets and the player picks. Passing `scope:sabotage_actor.var:next_versus_match` would be more precise, but **the same imprecision already exists in `6052`**, so tightening it would change vanilla behaviour in both events and felt like a separate fix. Happy to do it here if you'd prefer.

`make tiger`: `fatal: 0, error: 0, warning: 0`. Scope is Bug 1 only; Bugs 2–3 are still unlocated.

Worth verifying in-game: catch a saboteur in a **versus** contest (joust/duel/wrestling/board_game) and confirm he's out of *that* contest, his opponent advances, the contest still produces a winner, and he **still competes in the next contest**. Then repeat for a turn-based contest (archery/melee).

events/activities/tournaments/tournament_events.txt
```diff
@@ -14635,33 +14635,58 @@ tournament_events.6051 = { # Sabotage Outcome
...
 					}
 				}
-				remove_variable = contest_aptitude
-				add_character_flag = {
-					flag = disqualified_contest_flag #Unop: Use correct flag
-					months = 2
-				}
+				#Unop Only the aptitude was removed, which excludes the saboteur from turn-based contests but not from
+				#Unop versus brackets, where pairing ignores it, so he kept being matched and could reach the finals.
+				#Unop Route by phase to the proper disqualification effect, as the sibling event tournament_events.6052 does.
```
pharaox:
```suggestion
				#Unop Only the aptitude was removed, so the saboteur kept being matched and could reach the finals.
				# Route by phase to the proper disqualification effect, as the sibling event tournament_events.6052 does.
```

pharaox (commented):
**reviewer:** Looks good — this supersedes my own Review 1, which signed off on `remove_from_activity` and was wrong. @ProZeratul's objection was correct, and transplanting 6052's phase switch is the right fix.

The new block is byte-identical to 6052's (modulo indentation), including the `scope:activity = { switch = ... }` wrapper, and the enclosing `if` gate matches 6052's exactly.

Verified:
- **No double-application on the human-host path** — the main hazard I looked for. 6052's Disqualify option runs the switch *and then* fires 6051 on the saboteur (`:14914`), so 6051's new switch could have disqualified twice. It can't: 6051's qualified branch needs `exists = scope:sabotage_disqualified`, and both places that set it (`:14592`, `:14607`) are inside `scope:host = { if = { limit = { is_ai = yes } } }`. On the 6052 path the host is human by construction (`6050:14261`), so the scope is never set and the block is skipped.
- **Root cause confirmed independently.** `tournament_contest_versus_disqualification_effect` strips `contest_aptitude` *and* `progress_to_victory`, sets the flag, and `add_to_guest_subset`s `$MATCH$` into the next stage — advancing the opponent in the saboteur's place. That's the omitted step, and it also explains the "nobody was selected as winner" breakage: removing a bracket member without advancing anyone leaves the stage underpopulated.
- **Nothing lost on the versus path** — that effect writes its own `tournament_disqualified_log` *and* applies `tournament_disqualified_opinion`, so dropping the inlined copies avoids a real double-log.
- **On `MATCH`:** the imprecision is real but bounded — 6050's targets are restricted to `is_in_guest_subset = { name = qualified }` with an existing `progress_to_victory`, so `MATCH` is always a live qualified contestant and a legitimate participant advances. Identical in 6052, so mirroring it is not a regression. Flagging it for @ProZeratul rather than diverging silently was the right call.
- Both `else` branches correctly keep `remove_from_activity` for a *non-contestant* saboteur.
- `make tiger` re-run independently: 0 errors, 0 warnings.

**One Low finding:** on the **non-versus** path the fix drops behaviour `main` had — `tournament_contest_disqualification_effect` emits only a tooltip, so there's no log entry and, unmentioned in the writeup, no `tournament_disqualified_opinion` (-15) against the host for an AI saboteur. Not a regression versus the reference implementation (vanilla's own 6052 `fallback` has the identical gap), but sabotage does reach turn-based contests (`archery`/`falconry`/`horse_race` flavours), so it's worth knowing the AI no longer takes that opinion hit there.

**In-game test still needed:** catch a saboteur in a versus contest — confirm he's out of that contest, his opponent advances, the contest still yields a winner, and he still competes in the next contest. Then a turn-based contest, where the Low above applies.

ProZeratul:
I'll re-check this as I saw it going south quickly. Will merge when test is done

PR #589, opened 2026-07-19
Step China army command law down instead of resetting on Tension

### LINKED ISSUE #581: As China, Grand Expansionist Command reverts to Autonomous Command instead of Imperial Command on era change to tension
**Reported on Steam by `White Lich` on `2026-07-05 01:28`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> As China, when the Army Command structure is set as Grand Expansionist Command and the era changes to tension. The law reverts to the lowest level Autonomous Command instead of the next level down Imperial Command.

pharaox:
Fixes #581

**fixer:** As China, `celestial_army_liege_law_3` (Grand Expansionist Command) has a `can_keep` that requires the dynastic-cycle situation to be in the Stability/Expansion phase. When the era changes to Tension, that `can_keep` invalidates and the engine reverts the law to the group **default** (`celestial_army_liege_law_0`, Autonomous) — a drop to the floor instead of a graceful step down to Imperial Command (level 2).

Vanilla already solves the identical problem for `celestial_bureaucracy` by stepping it down one level (`celestial_bureaucracy_3` → `_2`) in the Stability/Expansion phase `on_end`. This fix mirrors that step-down for the army law: in the same `on_end` (both dynastic-cycle phase-group variants), if the China holder has `celestial_army_liege_law_3`, set `celestial_army_liege_law_2` before the engine's default-revert can fire.

Scope note: like the vanilla bureaucracy step-down, this steps the top-liege's law. The vassal-side cascade (`celestial_army_vassal_law_3`) and the exact engine timing (script step-down vs. the "revert within a month" auto-revert) should be confirmed in-game.

### DIFF
--- a/common/situation/situations/tgp_dynastic_cycle.txt
+++ b/common/situation/situations/tgp_dynastic_cycle.txt
@@ -694,6 +694,20 @@ dynastic_cycle = {
 						}
 						add_realm_law = celestial_bureaucracy_2
 					}
+					#Unop Step the army command law down one level instead of letting it revert to the group default (Autonomous), mirroring the bureaucracy step-down above
+					if = {
+						limit = {
+							has_realm_law = celestial_army_liege_law_3
+						}
+						add_realm_law = celestial_army_liege_law_2
+						#Unop Step the vassals' mirrored army law down too, so they don't cascade-revert to the group default (Autonomous) once the liege drops below level 3
+						every_vassal_or_below = {
+							limit = {
+								has_realm_law = celestial_army_vassal_law_3
+							}
+							add_realm_law_skip_effects = celestial_army_vassal_law_2
+						}
+					}
 				}
 			}
 


pharaox (commented):
**reviewer:** Needs work — the liege-side step-down is correct and at the right altitude (army_law_3 is legal only in Expansion, whose only successor is Instability), and it mirrors the shipped vanilla bureaucracy step-down. But the **vassal cascade** is unhandled.

- Medium: when the liege steps `celestial_army_liege_law_3 → _2`, vassals' `celestial_army_vassal_law_3.can_keep` (requires `liege has celestial_army_liege_law_3`) invalidates and reverts to the group default `celestial_army_vassal_law_0` — so the realm-wide 3→0 drop survives for vassals. Bureaucracy's liege-only precedent doesn't transfer: its default is `_1`, not the floor.
- Low: the second insertion is in the **Advancement** phase `on_end`, not Expansion — army_law_3 can't legally exist there (its can_keep requires Expansion), so it's an inert defensive mirror.

common/situation/situations/tgp_dynastic_cycle.txt
```diff
@@ -694,6 +694,13 @@ dynastic_cycle = {
 						}
 						add_realm_law = celestial_bureaucracy_2
 					}
+					#Unop Step the army command law down one level instead of letting it revert to the group default (Autonomous), mirroring the bureaucracy step-down above
+					if = {
+						limit = {
+							has_realm_law = celestial_army_liege_law_3
+						}
```
pharaox, line 701: **reviewer:** (Medium) This fixes the liege, but `celestial_army_vassal_law_3.can_keep` requires `liege = { has_realm_law = celestial_army_liege_law_3 }`. Once the liege is `_2`, every vassal's `vassal_law_3` invalidates and the engine reverts it to the group default `celestial_army_vassal_law_0` (the army vassal default is 0, unlike bureaucracy's `_1`). So vassals still collapse 3→0 — the realm-wide symptom the analysis flagged. Consider a parallel `celestial_army_vassal_law_3 → _2` step-down across affected vassals, or confirm in-game that vassals dropping to Autonomous is acceptable.

common/situation/situations/tgp_dynastic_cycle.txt
```diff
@@ -977,6 +984,13 @@ dynastic_cycle = {
...
```
pharaox: **reviewer:** (Low) This block is the **Advancement** phase `on_end` (`situation_dynastic_cycle_phase_stability_advancement`), not Expansion. `celestial_army_liege_law_3.can_keep` requires the Expansion phase, so the China holder can't legally hold `_3` here — this branch is almost certainly inert. Harmless (it mirrors the bureaucracy line, which *is* meaningful here since bureaucracy_3 is legal during Advancement), but the rationale is inaccurate.

pharaox:
**fixer:** Re-fix addressing the review's Medium finding (vassal cascade).

The liege step-down (`celestial_army_liege_law_3 → _2`) is retained. Added a parallel vassal step-down inside the same `if`: after the liege drops to `_2`, iterate `every_vassal_or_below` and step any holder of `celestial_army_vassal_law_3` to `celestial_army_vassal_law_2` via `add_realm_law_skip_effects` (matching the game_start precedent that sets vassal army laws with `add_realm_law_skip_effects = celestial_army_vassal_law_2`). This prevents the realm-wide 3→0 collapse: without it, once the liege leaves level 3 every vassal's `celestial_army_vassal_law_3.can_keep` invalidates and the engine reverts each to the group default `celestial_army_vassal_law_0` (Autonomous).

Order is deliberate — the liege is set to `_2` first so the vassals' `celestial_army_vassal_law_2.can_keep` (which requires `liege = { has_realm_law = celestial_army_liege_law_2 }`) holds when they are set.

tiger clean. Load-bearing assumption unchanged: the Expansion-phase `on_end` scripted step-down runs before the engine's "revert invalidated law to default within a month", corroborated by the shipped vanilla bureaucracy step-down in the same block. **Needs in-game verification** that both liege and all vassals land on Imperial (level 2), not Autonomous, on the transition to Tension.

pharaox:
`@fixer` Also check and address the second reviewer finding, even though it's Low.

pharaox:
**fixer:** Re-fix addressing the review's Low finding (per @pharaox's request).

Removed the second army step-down block — the one in `situation_dynastic_cycle_phase_stability_advancement.on_end`. It was inert dead code: `celestial_army_liege_law_3.can_keep` requires the China holder to be in `situation_dynastic_cycle_phase_stability_expansion`, so the holder can never legally hold `_3` during the Advancement phase, and its `if` limit (`has_realm_law = celestial_army_liege_law_3`) could never be true there. The vanilla `celestial_bureaucracy_3 → _2` step-down in that block is untouched, since bureaucracy `_3` *is* legal outside the two instability phases.

The Medium finding (vassal cascade) remains handled by the retained Expansion-phase block: liege `_3 → _2`, then `every_vassal_or_below` holding `celestial_army_vassal_law_3` stepped to `_2` (order deliberate so the vassals' `can_keep` sees the liege already at `_2`).

tiger clean. Still **needs in-game verification** that both liege and all vassals land on Imperial (2), not Autonomous (0), on the Expansion→Tension transition.

PR #590, opened 2026-07-19
Make Silk Road Steady Trading grant development growth percentage, not flat add

### LINKED ISSUE #586: Silk Road Steady Trading phase grants 0.01 Monthly Development instead of 1% Development Growth
**Reported on Steam by `White Lich` on `2026-07-15 06:06`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> For the silk road situation, in the Steady Trading phase, it grants 0.01 Monthly Development instead of 1% Development Growth for Everyone. As indicated in  https://ck3.paradoxwikis.com/Situation#The_Silk_Road .

pharaox:
Fixes #586

**fixer:** The Silk Road Steady Trading phase's subregion-wide `all` county_modifier used `development_growth = 0.01` (a flat +0.01 Monthly Development add) instead of `development_growth_factor = 0.01` (a +1% growth factor). The missing `_factor` is a vanilla typo: the parallel prosperity phase `phase_exceptional_bounty` uses `development_growth_factor = 0.02` in the identical `all` block (same "Very small boost to entire subregion" comment). Fixed to `development_growth_factor = 0.01`, matching the wiki's "1% Development Growth for Everyone".

Scoped strictly to the `all` block — the legitimate flat `development_growth = 0.01` in the `silk_road_domain` route blocks is untouched.

Adds a new Unop override of `common/situation/situations/tgp_silk_road.txt` (vanilla file added unmodified in its own commit first). Adding this override surfaced a latent, unrelated tiger scope false positive in `events/artifacts/artifact_events.txt` (a `var:banner_dynasty` comparison tiger can't type-infer; the value is a dynasty, so the comparison is correct); extended the existing `banner_dynasty` false-positive rule in `ck3-tiger.conf` to cover that file. `make tiger` is clean.

### DIFF
--- a/ck3-tiger.conf
+++ b/ck3-tiger.conf
@@ -216,6 +216,7 @@ filter = {
 				file = common/scripted_effects/00_decisions_effects.txt
 				file = events/court_maintenance_events.txt
 				file = events/dlc/ep1/ep1_fund_inspiration_events.txt
+				file = events/artifacts/artifact_events.txt
 			}
 		}
 		NAND = { # var:banner_dynasty/var:banner_house track scope:target which is documented polymorphic (title, house, or dynasty)
--- a/common/situation/situations/tgp_silk_road.txt
+++ b/common/situation/situations/tgp_silk_road.txt
@@ -1,0 +1,560 @@
+﻿
+silk_road_situation = {
+	illustration = "gfx/interface/illustrations/decisions/tgp_silk_road.dds"
+	situation_group_type = minor
+	gui_tooltip_group_focused = yes
+
+	#########################################################################
+	# Feature ownership is theft
+	# Design by the collective will of the design discipline
+	# Additional design by you, the people
+	#########################################################################
+
+	#window = the_silk_road
+	#gui_window_name = "window_the_silk_road"
+	#gui_participation_window_name = "window_situation_participation"
+	#map_mode = sub_regions
+
+	is_unique = yes
+	map_mode = silk_road
+	window = silk_road
+	gui_window_name = "window_silk_road"
+	##################################################
+	# Regions
+	##################################################
+	sub_regions = {
+		region_silk_road_proper_china = {
+			geographical_regions = { tgp_silk_road_china_region }
+			capital_province = 10148 # Chang'an (Zhouzhi)
+
+			map_color = { 0 168 107 }
+		}
+		region_silk_road_proper_tibet = {
+			geographical_regions = { tgp_silk_road_tibet_region }
+			capital_province = 9238 # Potala (Lhasa)
+
+			map_color = { 239 208 40 }
+		}
+		region_silk_road_proper_india = {
+			geographical_regions = { tgp_silk_road_india_region }
+			capital_province = 1362 # Lahur
+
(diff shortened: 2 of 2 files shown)


common/situation/situations/tgp_silk_road.txt
```diff
@@ -0,0 +1,560 @@
...
+			modifier_sets = {
+				silk_road_phase_effects = {
+					icon = "gfx/interface/icons/modifiers/goods_mixed.dds"
+
+					all = {
+						character_modifier = {
+							cultural_head_fascination_mult = 0.02
+						}
+						county_modifier = { # Very small boost to entire subregion - Main county effects are in on_start
+							development_growth_factor = 0.01 #Unop Was a flat development_growth add; the subregion-wide boost should be a percentage growth factor, matching the sibling phase_exceptional_bounty
```
pharaox, line 380: **reviewer:** Correct root-cause fix. `modifiers.log`: `development_growth` = flat monthly add, `development_growth_factor` = percentage — both legal, so tiger never caught the vanilla typo. This `all` block (subregion-wide, "for Everyone") now matches `phase_exceptional_bounty`/`phase_hardship`, which both use `development_growth_factor`. Nicely scoped: the legit flat `development_growth = 0.01` in the `silk_road_domain` route block (line 264) is left alone.

ck3-tiger.conf
```diff
@@ -216,6 +216,7 @@ filter = {
 				file = common/scripted_effects/00_decisions_effects.txt
 				file = events/court_maintenance_events.txt
 				file = events/dlc/ep1/ep1_fund_inspiration_events.txt
+				file = events/artifacts/artifact_events.txt
```
pharaox, line 219: **reviewer:** (Low) This suppression is *correct* — `var:banner_dynasty != scope:owner.dynasty` at `artifact_events.txt:253` is the documented-polymorphic banner-dynasty pattern this rule already covers. But the `## Fix` rationale that the Silk Road override "surfaced" it is inaccurate: I ran `make tiger` repeatedly and the warning is **non-deterministic** — it appears even with the Silk Road file removed (1/5 runs on an otherwise-main tree). It's a pre-existing flaky false positive, not introduced here. Keeping the rule is fine (it reliably clears the flakiness); just don't mistake it for override-caused. Worth noting so a one-shot "tiger is clean" run doesn't lead someone to drop the rule.

pharaox (commented):
**reviewer:** Looks good — the core fix is correct, minimal, and precisely scoped. `modifiers.log` confirms `development_growth` is a flat add and `development_growth_factor` a percentage; the `phase_steady_trading` `all` block was the lone outlier using the flat tag where both sibling phases use the factor. Diff vs vanilla is one line; the legit domain flat add (line 264) is untouched. Commit hygiene clean, `#Unop` accurate.

- Low: the `ck3-tiger.conf` addition is a *correct* suppression, but the claim that the Silk Road override surfaced the `var:banner_dynasty` warning is wrong — I verified it's a pre-existing, non-deterministic false positive that appears even without this override. Not a blocker.
- Low: untested in-game; confirm the tooltip reads "1% Development Growth".

PR #594, opened 2026-07-24
Fix flag-from-variable type error saving movement_petition_effect

### LINKED ISSUE #591: Internal issue
```
[18:53:16][E][jomini_scriptvalue.cpp:1485]: Value of wrong type in 'file: events/decisions_events/tgp_decision_events.txt line: 494 (tgp_decision_events.0100:after)'. Got value of type 'flag'
```

From

```
    after = {
        if = { #Unop save missing scope:movement_petition_effect for petition types that bypass movement_petition_change_laws_decision
            limit = { NOT = { exists = scope:movement_petition_effect } }
            save_scope_value_as = { name = movement_petition_effect value = var:movement_petition }
        }
```

@pharaox 
So you cannot transfer a variable with a flag to a scope.
But you can transfer a scope with a flag to a variable.

pharaox:
Fixes #591

**fixer:** The `tgp_decision_events.0100.after` block filled the missing `scope:movement_petition_effect` with `save_scope_value_as = { value = var:movement_petition }`. The `value` slot is parsed as a scriptvalue, and a flag read out of a variable can't be evaluated there — throwing the logged `jomini_scriptvalue.cpp:1485 ... Got value of type 'flag'` error and leaving the scope unset.

A flag *literal* is accepted by `save_scope_value_as` (vanilla's own `movement_petition_change_laws_decision` does exactly that), but a flag pulled from a variable is not. The scope is re-derived via a `switch` over `var:movement_petition` that saves the matching flag as a literal per case, covering all 15 petition flags the downstream `switch` (`grant_movement_petition_tooltip_effect`) handles. All 15 are required: `ai_movement_petition_change_laws_decision` bypasses `movement_petition_change_laws_decision` entirely and reaches event 0100 with any petition flag but no saved scope. The `NOT = { exists = scope:movement_petition_effect }` guard is kept, so the vanilla decision path (which already sets the scope from a literal) is untouched.

The switch is extracted into `unop_save_movement_petition_effect` (`common/scripted_effects/00_unop_effects.txt`), leaving the `.after` block a single call.

tiger clean.

### DIFF
--- a/common/scripted_effects/00_unop_effects.txt
+++ b/common/scripted_effects/00_unop_effects.txt
@@ -739,3 +739,30 @@ unop_distribute_determined_court_position_effect = {
 		}
 	}
 }
+
+# Saves scope:movement_petition_effect from var:movement_petition for petition types
+# that reach tgp_decision_events.0100 without it being set (e.g. via ai_movement_petition_change_laws_decision).
+# save_scope_value_as cannot take a flag read out of a variable, so the flag is re-derived as a literal per case.
+unop_save_movement_petition_effect = {
+	if = {
+		limit = { NOT = { exists = scope:movement_petition_effect } }
+		switch = {
+			trigger = var:movement_petition
+			flag:change_bureaucracy_laws_increase = { save_scope_value_as = { name = movement_petition_effect value = flag:change_bureaucracy_laws_increase } }
+			flag:change_bureaucracy_laws_decrease = { save_scope_value_as = { name = movement_petition_effect value = flag:change_bureaucracy_laws_decrease } }
+			flag:change_army_laws_increase = { save_scope_value_as = { name = movement_petition_effect value = flag:change_army_laws_increase } }
+			flag:change_army_laws_decrease = { save_scope_value_as = { name = movement_petition_effect value = flag:change_army_laws_decrease } }
+			flag:change_province_metropolitan = { save_scope_value_as = { name = movement_petition_effect value = flag:change_province_metropolitan } }
+			flag:change_province_industrial = { save_scope_value_as = { name = movement_petition_effect value = flag:change_province_industrial } }
+			flag:change_province_military = { save_scope_value_as = { name = movement_petition_effect value = flag:change_province_military } }
+			flag:change_province_protectorate = { save_scope_value_as = { name = movement_petition_effect value = flag:change_province_protectorate } }
+			flag:hold_examinations = { save_scope_value_as = { name = movement_petition_effect value = flag:hold_examinations } }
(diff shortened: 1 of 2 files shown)


pharaox:
`@fixer` Are all of these required, as the original code talks about "petition types that bypass movement_petition_change_laws_decision"? If yes, extract this new code as a `unop_` effect as it's quite volumnous.

pharaox:
**fixer:** Yes, all 15 are required. The separate `ai_movement_petition_change_laws_decision` (`tgp_dynastic_cycle_decisions.txt` ~l.1909–2051) fully bypasses `movement_petition_change_laws_decision`: it sets `var:movement_petition` to *any* of the 14 real petition flags and fires event 0100 **without** ever saving `scope:movement_petition_effect`. Plus `change_candidate_score_law`/`hold_examinations` reach 0100 directly from the first decision (l.836–890) unsaved. The consumer `switch` (`grant_movement_petition_tooltip_effect`) matches all 15, so the save has to cover all 15 too.

Extracted the switch into `unop_save_movement_petition_effect` in `common/scripted_effects/00_unop_effects.txt` (guard included). The `.after` block is now a one-liner. tiger clean.

pharaox (commented):
**reviewer:** Looks good — Fix 2 resolves the type error correctly and fully addresses the earlier feedback.

- **Root cause:** the flag-from-variable `save_scope_value_as` is replaced by a `switch` saving each flag as a **literal** (`value = flag:X`) — the type-correct form, matching vanilla `tgp_dynastic_cycle_decisions.txt` l.1279–1282, and producing exactly the flag value-scope the consumer `grant_movement_petition_tooltip_effect` switches on.
- **Enumeration exact:** diffed the 15 `flag:` cases against the consumer switch — 15/15 exact match, no missing/extra case.
- **"Are all required?" — vetted yes:** `ai_movement_petition_change_laws_decision`'s `set_variable`-only block and the first decision's `change_candidate_score_law`/`hold_examinations` paths fire event 0100 with `var:movement_petition` set but **no** `save_scope_value_as`, so the scope arrives unset for all 15.
- **"Extract as unop_ effect" — done:** `unop_save_movement_petition_effect` in the dedicated `00_unop_effects.txt`, `.after` reduced to a one-liner. Guard `NOT = { exists = scope:movement_petition_effect }` preserved so the vanilla path is untouched.
- `#Unop` marker retained; `make tiger` clean; no incidental whitespace (pre-existing missing trailing newline preserved).

Worth an in-game check: previously-broken bypass petition types (esp. via `ai_movement_petition_change_laws_decision`) now show the correct tooltip and error.log is clean.

PR #596, opened 2026-07-24
Fix marshal overtime boost/decrease scales reading wrong scope and variable

### LINKED ISSUE #593: marshal_overtime_boost_scale & marshal_overtime_decrease_scale have issues
`marshal_overtime_boost_scale` is checking councillor_liege but is using liege directly to add the variable

Image

`marshal_overtime_decrease_scale` have the same scoping issue but also is checking for the `marshal_overtime_boost` variable but is using `marshal_overtime_decrease`

pharaox:
Fixes #593

**fixer:** Both `marshal_overtime_boost_scale` and `marshal_overtime_decrease_scale` are consumed in `council_owner_modifier` blocks of `task_manage_guards`, where root = the council owner. The overtime variables are stored on the council owner (= `scope:councillor_liege`), but the scales read them via `liege` — the owner's *own* liege, a different character with no such variable — so the ramp was read as 0.

Two fixes:
- **Wrong scope (both scales):** `add = liege.var:marshal_overtime_boost` / `...decrease` → `add = scope:councillor_liege.var:...`, matching the `else_if` guards that already test `scope:councillor_liege`.
- **Wrong variable (decrease scale):** its `else_if` guarded on `has_variable = marshal_overtime_boost` but adds `marshal_overtime_decrease` — a copy-paste error. Guard now tests `marshal_overtime_decrease`.

Adds `99_marshal_values.txt` as a new vanilla override (unmodified in the first commit, fix in the second). tiger clean. Vanilla bug.

### DIFF
--- a/common/script_values/99_marshal_values.txt
+++ b/common/script_values/99_marshal_values.txt
@@ -1,0 +1,663 @@
+﻿
+######################
+# Organize Levies
+######################
+
+### Improving Levies
+# Base value
+marshal_organize_levies_base = {
+	add = 1
+}
+
+marshal_organize_levies_skill_base = {
+	add = martial
+	multiply = 2
+}
+
+# Erudition bonus
+marshal_organize_levies_erudition_bonus = {
+	if = {
+		limit = { councillor_liege_has_erudition_legacy_5_perk = yes }
+		add = {
+			value = marshal_organize_levies_base
+			add = marshal_organize_levies_skill_base
+			multiply = {
+				add = erudition_legacy_5_percentage
+				divide = 100
+			}
+		}
+	}
+}
+
+# Tradition: Family business
+marshal_organize_levies_family_business_bonus = {
+	if = {
+		limit = { councillor_liege_has_family_business_bonus = yes }
+		add = {
+			value = marshal_organize_levies_base
+			add = marshal_organize_levies_skill_base
+			multiply = {
+				add = family_business_councillors_percentage
+				divide = 100
+			}
+		}
+	}
+}
+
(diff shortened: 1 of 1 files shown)


pharaox (commented):
**reviewer:** Looks good — both defects in the marshal-overtime scale pair are fixed correctly.

- **Bug A (scope):** verified the scales are consumed only in `task_manage_guards` `council_owner_modifier` blocks (root = council owner), and the variables are stored on the owner via `liege = { set_variable ... }` (root = councillor). So `scope:councillor_liege.var:...` is the correct read; the old `liege.var:...` reached the owner's own liege. Both call sites covered (grepped mod + vanilla).
- **Bug B (guard):** the decrease guard now tests `has_variable = marshal_overtime_decrease`, matching the variable read.
- Commit 1 adds the vanilla file byte-identical (verified with `diff`); commit 2 has exactly the three edits. `#Unop` markers accurate. `make tiger` clean.

Possible **separate** follow-up (not this PR's scope): `marshal_maa_overtime_boost_scale` / `marshal_knights_overtime_boost_scale` read `var:marshal_maa_overtime_boost` bare in a `council_owner_modifier` (root = owner) while that variable is stored bare on the councillor in `task_train_commanders`. It may read the wrong character, but unlike the overtime pair its guard and read are *consistent* (both bare `var:`), so it is not the same guard/read-mismatch bug — worth a dedicated investigation rather than folding into this PR.

Worth an in-game check per the reporter's screenshot: the knight-effectiveness / enemy-scheme-success modifiers now ramp with the tracked variable, and error.log no longer shows the missing-variable read.

PR #597, opened 2026-07-24
Fix accolade successor loop from low-prowess existing knights

### LINKED ISSUE #576: Accolade successor with too-low prowess is rejected as knight, causing repeated succession loop
**Reported on Steam by `jy` on `2026-06-30 13:37`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> an issue i seen reported in the forums, accolade successor has too low prowess to be a knight due to only the strong tradition, accolade knight is then rejected, triggering another succession, and problem repeats

pharaox:
Fixes #576

**fixer:** With the `only_the_strong` tradition (`minimum_prowess_for_knights`), knighthood requires `prowess >= 12`, but acclaimed-knight successor selection only requires `prowess >= 8`. `new_acclaimed_knight_search_trigger` does apply `can_be_knight_trigger`, but exempts characters who are *already a knight of the owner* (`trigger_if NOT is_knight_of`). An existing knight whose prowess is 8–11 (e.g. knighted before the culture took the tradition, or whose prowess later dropped) thus passes successor selection but is invalid as a knight, so gets rejected — triggering another succession, and looping.

Fix: enforce the knight prowess minimum (`prowess >= 12` when the owner's culture has `minimum_prowess_for_knights`) unconditionally in `new_acclaimed_knight_search_trigger`, so already-knighted low-prowess characters are excluded from selection. This targets the real invariant rather than widening the existing exemption. All 20 selection paths funnel through this one trigger.

Adding the vanilla file as a new override surfaced 5 unrelated latent vanilla `strict-scopes` warnings (`fanatic_attribute_trigger` etc. reference `scope:owner`, guarded by `?=`); flagged as false positives in `ck3-tiger.conf`.

### DIFF
--- a/common/scripted_effects/00_accolades_scripted_effects.txt
+++ b/common/scripted_effects/00_accolades_scripted_effects.txt
@@ -1,0 +1,10511 @@
+﻿##################################################
+# Accolade Effects
+
+accolades_development_growth_for_least_developed_county_effect = {
+	# First, should this apply at all?
+	if = {
+		limit = {
+			# Do you have a knight with this accolade at all?
+			any_active_accolade = {
+				OR = {
+					has_accolade_parameter = accolade_development_growth_on_tournament_completion_low
+					has_accolade_parameter = accolade_development_growth_on_tournament_completion_high
+				}
+			}
+			# Plus, pre-filter out landless rulers.
+			is_landed_or_landless_administrative = yes
+		}
+		# Now, find our least developed county.
+		ordered_held_title = {
+			title_tier = county
+			order_by = {
+				value = development_level
+				multiply = -1
+			}
+			save_scope_as = least_developed_county
+		}
+		# Next, we'll need to apply a bonus to it.
+		## Low boost.
+		if = {
+			limit = {
+				any_active_accolade = {
+					accolade_parameter = accolade_development_growth_on_tournament_completion_low
+				}
+			}
+			scope:least_developed_county = { change_development_progress_with_overflow = accolade_development_growth_on_tournament_completion_value }
+		}
+		## High boost.
+		if = {
+			limit = {
+				any_active_accolade = {
+					accolade_parameter = accolade_development_growth_on_tournament_completion_high
+				}
+			}
+			scope:least_developed_county = { change_development_level = accolade_development_level_on_tournament_completion_value }
+		}
+	}
+}
+
+accolades_contender_tournament_xp_gain_effect = {
+	custom_tooltip = accolade_tournament_xp_gain_contender_only.tt
+	hidden_effect = {
+		if = {
+			limit = {
+				NOT = { has_trait = tourney_participant }
+			}
+			add_trait = tourney_participant
+		}
+	}
+	# Low level bonus
+	if = {
+		limit = {
+			accolade = {
+				has_accolade_parameter = accolade_tournament_xp_gain_contender_only_low
+			}
+		}
+		
+		hidden_effect = {
+			random_list = {
+				10 = {
+					show_chance = no
+					add_trait_xp = {
+						trait = tourney_participant
+						track = foot
+						value = accolade_contender_trait_xp_low_value
+					}
+				}
+				10 = {
+					show_chance = no
+					add_trait_xp = {
+						trait = tourney_participant
+						track = horse
+						value = accolade_contender_trait_xp_low_value
+					}
+				}
+				10 = {
+					show_chance = no
+					add_trait_xp = {
+						trait = tourney_participant
+						track = bow
+						value = accolade_contender_trait_xp_low_value
+					}
+				}
+				10 = {
+					show_chance = no
+					add_trait_xp = {
+						trait = tourney_participant
+						track = wit
+						value = accolade_contender_trait_xp_low_value
+					}
+				}
+			}
+		}
+	}
+	# Mid level bonus
+	else_if = {
+		limit = {
+			accolade = {
+				has_accolade_parameter = accolade_tournament_xp_gain_contender_only
+			}
+		}
+		hidden_effect = {
+			random_list = {
+				10 = {
+					add_trait_xp = {
+						trait = tourney_participant
+						track = foot
+						value = accolade_contender_trait_xp_medium_value
+					}
+				}
+				10 = {
+					add_trait_xp = {
+						trait = tourney_participant
+						track = horse
+						value = accolade_contender_trait_xp_medium_value
+					}
+				}
+				10 = {
+					add_trait_xp = {
+						trait = tourney_participant
+						track = bow
+						value = accolade_contender_trait_xp_medium_value
+					}
+				}
+				10 = {
+					add_trait_xp = {
+						trait = tourney_participant
+						track = wit
+						value = accolade_contender_trait_xp_medium_value
+					}
+				}
+			}
+		}
+	}
+	# High level bonus
+	else_if = {
+		limit = {
+			accolade = {
+				has_accolade_parameter = accolade_tournament_xp_gain_contender_only_high
+			}
+		}
+		hidden_effect = {
+			random_list = {
+				10 = {
+					add_trait_xp = {
+						trait = tourney_participant
+						track = foot
+						value = accolade_contender_trait_xp_high_value
+					}
+				}
+				10 = {
+					add_trait_xp = {
+						trait = tourney_participant
+						track = horse
+						value = accolade_contender_trait_xp_high_value
+					}
+				}
+				10 = {
+					add_trait_xp = {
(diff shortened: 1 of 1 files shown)


ck3-tiger.conf
```diff
@@ -127,6 +127,16 @@ filter = {
 				file = events/scheme_events/murder_scheme/assassination_ongoing_events.txt
 			}
 		}
+		NAND = { # accolade attribute triggers guard scope:owner with `?=`; owner is intentionally optional in accolade-fit and knight-qualify contexts
```
pharaox:
```suggestion
		NAND = { # owner is used extensively in accolade contexts, so we assume it exists, although undocumented
```

common/scripted_triggers/04_ep2_accolade_triggers.txt
```diff
@@ -0,0 +1,1575 @@
...
+	trigger_if = {
+		limit = {
+			NOT = {
+				is_knight_of = scope:owner
+			}
+		}
+		can_be_knight_trigger = { ARMY_OWNER = scope:owner }
+	}
+	#Unop An already-knighted low-prowess character is exempted from can_be_knight_trigger above,
+	#Unop so enforce the knight prowess minimum here to avoid selecting an invalid acclaimed successor.
```
pharaox:
```suggestion
	#Unop Enforce the knight prowess minimum here to avoid selecting an invalid acclaimed successor.
```

common/scripted_triggers/04_ep2_accolade_triggers.txt
```diff
@@ -0,0 +1,1574 @@
...
+
+	trigger_if = {
+		limit = {
+			NOT = {
+				is_knight_of = scope:owner
+			}
+		}
+		can_be_knight_trigger = { ARMY_OWNER = scope:owner }
+	}
+	#Unop Enforce the knight prowess minimum here to avoid selecting an invalid acclaimed successor.
```
pharaox: **reviewer:** This guard is correct and faithfully mirrors vanilla `can_be_knight_trigger`, but it only covers *selection*. Prowess is mutated after this point: `accolade_squires_boost_prowess_effect` can apply `add_prowess_skill = -1` for pathetic-quality squires, and its trailing floor only raises `prowess < 8` to 8 — so a candidate that passes here at exactly 12 can be knighted at 11.

The same effect is also the only prowess normalisation applied to characters from `accolade_character_creation_effect`, which bypasses this trigger entirely. Making that floor culture-aware would fix both paths and make this block redundant.

pharaox (commented):
**reviewer:** Needs work — the selection guard is correct and well-placed, but two downstream vanilla paths can still produce a sub-12-prowess acclaimed knight, so the loop can still be seeded.

- **High:** `accolade_squires_boost_prowess_effect` runs *after* selection and *before* `set_knight_status = force` (`00_accolades_scripted_effects.txt` 10297 → 10331). Pathetic-quality squires apply `add_prowess_skill = -1`, and the corrective floor at the end of that effect only lifts `prowess < 8` up to **8**, not 12. A candidate selected at exactly prowess 12 can be knocked to 11 and then knighted — the exact invalid-knight state this PR targets. That floor is hardcoded to the successor threshold (8), not the knight threshold, so it is tradition-unaware.
- **Medium:** the `accolade_character_creation_effect` fallback (10255/10290) never runs `new_acclaimed_knight_search_trigger` at all. Its templates use `random_traits = yes` with no prowess guarantee and then hit the same `< 8` floor, so it can produce a prowess 8–11 acclaimed knight under the tradition. Narrowing the candidate pool makes this path *more* likely to be taken.

A culture-aware floor (raise to 12 when the owner culture has `minimum_prowess_for_knights`) would cover both paths and would subsume the selection-side guard.

Verified clean: all 20 call sites funnel through this one trigger (no missed sites); `make tiger` re-run independently reports 0 warnings; the 4 `ck3-tiger.conf` suppressions are genuine `?=`-guarded false positives; and the vanilla file in `9d8df33e` is byte-identical to the 1.19 base-game file.

The in-game test is still needed — the High finding is a concrete scenario where the loop would not stop.

pharaox:
**fixer:** Reworked the fix to the altitude suggested in review — the branch history was rewritten, so the previous approach is gone.

**What changed:** instead of guarding *selection* in `new_acclaimed_knight_search_trigger`, the fix now makes the prowess floor in `accolade_squires_boost_prowess_effect` culture-aware: when the owner's culture has `minimum_prowess_for_knights`, the floor is 12 instead of 8.

This resolves both review findings, because that floor is the single chokepoint on the path — it runs on `scope:chosen_knight` (selected **or** created by `accolade_character_creation_effect`) at `00_accolades_scripted_effects.txt:10297`, immediately before `set_knight_status = force` at 10331:

- **High** (post-selection prowess mutation): the floor now runs *after* `add_prowess_skill = -1` for pathetic squires, so a candidate knocked to 11 is raised back to 12.
- **Medium** (created-character fallback): created characters go through the same floor, so they can no longer be knighted at 8–11.

The selection-side guard is therefore redundant and was dropped, which also removes the `04_ep2_accolade_triggers.txt` override and its `strict-scopes` suppression entirely — the mod no longer overrides that file, and `ck3-tiger.conf` is untouched.

Overriding `00_accolades_scripted_effects.txt` surfaced 4 latent vanilla `scopes` warnings: `location_has_harsh_winter_trigger` is called inside `any_county_province`, but `location` only accepts character/army/combat (per `event_targets.log`), so it never resolves and the harsh-winter squire-trait branch silently never matches. Fixed in place with the direct `has_province_modifier = winter_harsh_modifier` rather than suppressed, since it's a real bug.

`make tiger`: `fatal: 0, error: 0, warning: 0`. Still needs the in-game check that the succession loop actually stops.

common/scripted_triggers/04_ep2_accolade_triggers.txt
pharaox: **fixer:** Agreed — done. The floor in `accolade_squires_boost_prowess_effect` is now culture-aware (12 under `minimum_prowess_for_knights`), which covers both the post-selection `add_prowess_skill = -1` mutation and the `accolade_character_creation_effect` path. As you noted, that makes this selection guard redundant, so the whole `04_ep2_accolade_triggers.txt` override is dropped along with its tiger suppression.

pharaox (commented):
**reviewer:** Looks good — the rework lands at the right altitude and closes both Review 1 findings.

Making the existing prowess floor culture-aware, instead of guarding selection, is the higher-level fix: it is the single chokepoint both the selected and the created successor pass through, immediately before `set_knight_status = force`. Dropping the `04_ep2_accolade_triggers.txt` override (and its tiger suppression with it) is a real simplification — one file touched, `ck3-tiger.conf` untouched.

Verified:
- **Chokepoint holds.** `accolade_create_squire_effect` is the sole caller of `accolade_squires_boost_prowess_effect`, reached from the only appointment path in the game (`on_accolade_create_squire`); no other `set_accolade_successor` call sites exist in vanilla. Both paths converge on `scope:chosen_knight` (10326) before `set_knight_status = force` (10363).
- **Ordering correct.** The floor runs after the pathetic-squire `add_prowess_skill = -1` and after `accolade_squires_boost_martial_tourney_effect`, and nothing after it lowers prowess — the "selected at 12, knighted at 11" scenario is closed.
- **Created path closed.** The templates are age 16–22, `faith = faith`, non-clergy, healthy, so prowess was their only `can_be_knight_trigger` blocker; they now hit the same floor.
- **`scope:owner` guaranteed** by `save_scope_as = owner` at the top of `accolade_create_squire_effect`, and it is the same character passed as `ARMY_OWNER` to `can_be_knight_trigger`. `?=` falls back to the vanilla `< 8` branch. The new branch is a faithful clone of that branch (8→12), correctly wired as `if`/`else_if`.
- **Residual loop risk is not prowess** — the search trigger already enforces `is_clergy = no`, `is_healthy = yes` and full `can_be_knight_trigger` for non-knights.
- **Harsh-winter fix correct and in scope.** `location` resolves only on character/army/combat (`event_targets.log:1695`), so inside `any_county_province` the branch was dead. Pattern class enumerated: all 4 province-scope misuses fixed; the other 27 call sites across vanilla are character-scope and correct. Fixed rather than suppressed, per CLAUDE.md.
- `make tiger` re-run independently: 0 errors, 0 warnings. Vanilla file in `d4d99959` is byte-identical to 1.19 and added unmodified in its own commit.

Note: the earlier inline comments on this PR are stale — they target the Fix 1 files the branch rewrite removed.

**In-game test still required:** confirm the loop actually stops with `only_the_strong`. The engine re-trigger stays invisible in script, so review can't settle that one.

PR #598, opened 2026-07-24
Let Japanese independent regent form a cadet house when house head is elsewhere

### LINKED ISSUE #560: create_cadet_branch_decision unconditionally blocks Japanese independent admin regent from forming a cadet house
**Reported on Steam by `jy` on `2026-06-22 07:16`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> as per the previous comment,  "it has been 20+ years that he did not try to create a cadet house", that is because under `create_cadet_branch_decision`,
>
> modifier = { # Ruler should not break family apart
> 		government_is_japanese_trigger = yes
> 		is_independent_ruler = yes
> 		add = -1000
> 		}
>
> it blocks the japanese admin regent from forming cadet house unconditionally even if they do not have house member within the realm
>
> basically inheriting japan realm to a new house via matri-marry shenanigans breaks the japanese govt,
>
> compared that to celestial empire which doesnt cause any significant issue, hegemon still form cadet house and properly get a domicile & family title

pharaox:
Fixes #560

**fixer:** In `create_cadet_branch_decision`'s `ai_will_do`, the `# Ruler should not break family apart` modifier applied `-1000` to *every* independent Japanese ruler (`government_is_japanese_trigger = yes` + `is_independent_ruler = yes`), with no family-relationship condition. For an independent Japanese admin regent whose realm was inherited by a foreign house, this unconditional penalty cancels the intended `+1000` booster ("admin top liege with house head in a different realm should create a cadet branch ASAP"), so the AI never reliably forms the cadet house and the Japanese government stays broken. Celestial Empire has no such penalty and works fine.

Fix: gate the penalty on `top_liege = house.house_head.top_liege`, so it no longer fires when the ruler's house head is in a different realm — the exact case the `+1000` booster targets. This is vanilla's own idiom for "house head is in a different realm" (see the `is_valid` admin branch). When the family *is* in-realm, the penalty still applies as before.

### DIFF
--- a/common/decisions/00_dynasty_decisions.txt
+++ b/common/decisions/00_dynasty_decisions.txt
@@ -1,0 +1,320 @@
+﻿### Create Cadet Branch ###
+create_cadet_branch_decision = {
+	picture = {
+		trigger = {
+			government_is_japanese_trigger = yes
+		}
+		reference = "gfx/interface/illustrations/decisions/tgp_poet.dds"
+	}
+	picture = {
+		reference = "gfx/interface/illustrations/decisions/decision_dynasty_house.dds"
+	}
+
+	decision_group_type = major
+	ai_check_interval_by_tier = {
+		barony = 0
+		county = 6
+		duchy = 6
+		kingdom = 6
+		empire = 6
+		hegemony = 6
+	}
+
+	is_shown = {
+		is_playable_character = yes
+		exists = house.house_head
+		is_house_head = no
+		highest_held_title_tier > tier_barony
+		NOR = {
+			government_has_flag = government_is_holy_order
+			government_has_flag = government_is_theocracy
+			AND = {
+				faith = { has_doctrine = doctrine_theocracy_temporal }
+				has_council_position = councillor_court_chaplain
+			}
+			AND = {
+				has_tgp_dlc_trigger = yes
+				tgp_is_in_ceremonial_house_trigger = yes
+			}
+		}
+	}
+
+	is_valid = {
+		is_adult = yes
+		government_allows = create_cadet_branches
+		# Must have living children of same house
+		custom_tooltip = {
+			text = create_cadet_branch_decision_living_children
+			any_child = {
+				count >= 1
+				is_alive = yes
+				house ?= root.house
+			}
+		}
+		# No direct descendant is liege or above
+		custom_tooltip = {
+			text = create_cadet_branch_decision_liege_is_descendant
+			any_liege_or_above_is_descendant = no
+		}
+		# If you are Clan, you cannot be in the same realm as your House Head. For simplicity, we'll just force you to be independent.
+		trigger_if = {
+			limit = {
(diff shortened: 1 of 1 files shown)


common/decisions/00_dynasty_decisions.txt
```diff
@@ -0,0 +1,321 @@
...
+			add = 50
+		}
+
+		modifier = { # Prefer to branch away from house head
+			government_is_japanese_trigger = yes
+			NOT = { is_close_or_extended_family_of = house.house_head }
+			add = 50
+		}
+
+		modifier = { #Unop Don't penalize breaking apart a family that isn't in the ruler's realm
```
pharaox: **reviewer:** *Low / non-blocking.* This replaces the vanilla intent comment `# Ruler should not break family apart` rather than preserving it. The project leans toward keeping vanilla text and marking the change — and the `#Unop` note on the added line below already carries the rationale, so the original could stay:

```suggestion
		modifier = { # Ruler should not break family apart
```

Purely cosmetic — the replacement text is accurate and arguably clearer. Flagging only because the vanilla comment recorded the modifier's original design intent, which is now recoverable only from the base-game file.

pharaox (commented):
**reviewer:** Looks good — minimal, correctly targeted, and it reuses vanilla's own expression of the condition instead of inventing a parallel one.

What I verified beyond the diff:

- **The load-bearing assumption is corroborated, not asserted.** `top_liege = house.house_head.top_liege` is the *identical* predicate vanilla uses in this same decision's `is_valid` (line 72), with the comment "But we do want the AI to create a cadet branch ASAP if the house head is in a different realm". The fix mirrors a verified mechanism.
- **No other guard re-blocks the case.** `is_shown` passes; `is_valid` routes through the administrative branch (`japan_administrative_government` has `administrative = yes` and `create_cadet_branches = yes`), and the AI-only residual conditions hold for a regent married into a foreign house. So the fix is not a no-op.
- **The `+1000` booster is reachable and no longer cancelled** — the added condition makes it and the `-1000` mutually exclusive by construction. The sibling `-1000` just below requires a close/extended-family or noble-family-title link to the house head, which does not hold here.
- **Pattern class enumerated.** The three Japanese cadet-branch variants in `tgp_japan_decisions.txt` were each checked; none carries an unconditional `is_independent_ruler` `-1000`. The outlier is unique to this decision, so no reproducing sibling was left untouched.
- **No side effects.** The `00_religious_interactions.txt` call sites use `execute_decision`/`can_execute_decision`, which consult `is_shown`/`is_valid`, not `ai_will_do`. Player-facing availability is unchanged.
- `make tiger` re-run independently: 0 warnings. Vanilla file in `22951972` is byte-identical to 1.19 and added in its own commit.

Only a `Low` note (inline). Worth an in-game check that the regent actually takes the decision within a reasonable span, since AI timing depends on the aggregate weight.

PR #599, opened 2026-07-24
Let a settling nomad GOK reform reach the Meritocratic Khanate government

### LINKED ISSUE #552: Reform Great Khanate decision for GOK lacks option to convert to Meritocratic Khanate
**Reported on Steam by `jy` on `2026-06-26 04:44`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> reform great khanagate decision for GOK doesnt have option to covert to meritocratic khanate, only meritocratic (doesnt seem consistent lore-wise)
>
> there is a workaround though, after clicking on the decision to reform, while game paused, click on the decision to become meritocratic khanate immediately, then the next game day reform events will then be triggered, becoming a meritocratic khanate hegemony
>
> there is several side effects though, feels like multiple bugs involved, mostly lost out on the free treasury and cash

pharaox:
Fixes #552

**fixer:** The "Reform Great Khanate" reform (`mpo_gok_world_conquest_decision` → event `mpo_greatest_of_khans.0030` option b) converts a GOK to administration via `convert_to_administrative_from_feudalism_effect` with `flag:dynamic`, whose government is resolved in `change_to_administrative_effect`. In that resolver, the **Meritocratic Khanate** (`steppe_admin_government`) branch is the only administrative sub-government lacking an independent-ruler resolution path — Ritsuryo, Celestial, and Meritocratic all have a `trigger_else_if` for the independent case, but the steppe_admin branch only fires when the ruler is already a vassal of a `steppe_admin` liege (the selected-government case is deliberately `always = no`). So an independent settling steppe nomad can never reach the Meritocratic Khanate via the reform and instead skips to plain Meritocratic/Administrative — despite Meritocratic Khanate being the natural steppe-admin form.

Fix: add an independent-ruler `trigger_else_if` to the steppe_admin branch (evaluated before the Meritocratic branch), mirroring the sibling branches, gated on a new `unop_should_become_steppe_admin_trigger` (independent nomad with a steppe capital). Restricted to current nomads so the many other callers of this shared effect are unaffected.

Adding `00_administrative_effects.txt` as a new override surfaced two unrelated latent tiger warnings, both handled here: a `treasury = 0` "exactly equal" logic warning (deliberate vanilla check — flagged false positive) and, via the `on_death` call path, a real unset-`scope:great_yuan_title` bug in `shift_counties_on_mongol_succession_effect` — a second occurrence of a block already Unop-fixed elsewhere in that effect, now fixed to use `title:e_great_yuan` directly.

Reads on the bug/enhancement line: this adds no player-facing option or new government, only restores the resolver symmetry so an existing dynamic destination becomes reachable. **Needs in-game verification** with a settling nomad GOK reforming.

### DIFF
--- a/common/scripted_effects/00_administrative_effects.txt
+++ b/common/scripted_effects/00_administrative_effects.txt
@@ -195,6 +195,12 @@ change_to_administrative_effect = {
 					# There's a separate decision to go from Nomad to Meritocratic Khanate, so you cannot pick it from here
 					always = no
 				}
+				trigger_else_if = { #Unop An independent settling nomad should reach the Meritocratic Khanate, not skip it for plain Meritocratic
+					limit = {
+						scope:governor_liege ?= this
+					}
+					unop_should_become_steppe_admin_trigger = yes
+				}
 				trigger_else = {
 					scope:governor_liege = { government_has_flag = government_is_steppe_admin }
 				}
--- a/common/scripted_effects/00_administrative_effects.txt
+++ b/common/scripted_effects/00_administrative_effects.txt
@@ -1,0 +1,550 @@
+﻿
+change_to_administrative_hereditary_effect = {
+	scope:recipient = {
+		save_scope_as = petition_vassal
+		every_held_title = {
+			limit = {
+				tier = scope:petition_vassal.highest_held_title_tier
+				is_landless_type_title = no
+				is_noble_family_title = no
+			}
+			custom_tooltip = {
+				text = petition_liege_house_governorship_rights_tt
+				set_variable = {
+					name = petition_house_rights
+					value = scope:petition_vassal.house
+					years = 100
+				}
+			}
+		}
+	}
+}
+
+change_to_administrative_holding_effect = {
+	# Structured like this for clean tooltips
+	capital_province = {
+		if = {
+			limit = {
+				OR = {
+					has_holding_type = tribal_holding
+					has_holding_type = herder_holding
+					has_holding_type = nomad_holding
+				}
+			}
+			county.holder = { custom_tooltip = change_to_administrative_holding_tt }
+			hidden_effect = { set_holding_type = castle_holding }
+		}
+		county = {
+			if = {
+				limit = { has_county_modifier = mpo_siberian_permafrost_modifier }
+				holder = { custom_tooltip = change_to_administrative_permafrost_tt }
+				hidden_effect = { remove_county_modifier = mpo_siberian_permafrost_modifier }
+			}
+			else_if = {
+				limit = { has_county_modifier = mpo_siberian_permafrost_modifier_bad }
+				holder = { custom_tooltip = change_to_administrative_permafrost_bad_tt }
+				hidden_effect = { remove_county_modifier = mpo_siberian_permafrost_modifier_bad }
+			}
+		}
+	}
+}
+
+change_to_administrative_effect = {
+	save_scope_as = governor
+	if = {
+		limit = { exists = scope:new_liege }
+		scope:new_liege = { save_scope_as = governor_liege }
+	}
+	else = {
+		top_liege = { save_scope_as = governor_liege }
+	}
+	# exactly 0 treasury,  i.e. newly get this feature
+	if = {
+		limit = {
+			OR = {
+				has_treasury = no
+				treasury = 0
+			}
+		}
+		save_scope_as = new_treasury
+	}
+	# Set the correct government
+	if = {
+		limit = {
+			OR = {
+				NOT = { government_allows = administrative }
+				AND = {
+					government_allows = administrative
+					scope:governor_liege = { government_has_flag = government_is_japan_administrative }
+					NOT = { government_has_flag = government_is_japan_administrative }
+				}
+				AND = {
+					government_allows = administrative
+					scope:governor_liege = { government_has_flag = government_is_meritocratic }
+					NOT = { government_has_flag = government_is_meritocratic }
+				}
+				AND = {
+					government_allows = administrative
+					scope:governor_liege = { government_has_flag = government_is_celestial }
+					NOT = { government_has_flag = government_is_celestial }
+				}
+				AND = {
+					government_allows = administrative
+					scope:governor_liege = { government_has_flag = government_is_steppe_admin }
+					NOT = { government_has_flag = government_is_steppe_admin }
+				}
+			}
+		}
+		# Only case where Estate type is shared between Admin and non-Admin
+		if = {
+			limit = {
+				NAND = {
+					government_has_flag = government_is_japan_feudal
+					scope:governor_liege = { government_has_flag = government_is_japan_administrative }
+				}
+			}
+			save_scope_as = new_admin
+		}
+		# If we are greek, let's ensure we become administrative
+		if = {
+			limit = {
+				NOT = { exists = scope:has_selected_a_government }
+				scope:governor_liege = {
+					this = scope:governor # not vassal of some other admin liege
+					culture = {
+						OR = {
+							this = culture:greek
+							any_parent_culture_or_above = { this = culture:greek }
+						}
+					}
+				}
+			}
+			save_scope_value_as = {
+				name = has_selected_a_government
+				value = flag:administrative
+			}
+		}
+		# Change government
+		# We first check if you have a selected government, then if you are independent, then what your liege has
+		if = { # Ritsuryo
+			limit = {
+				trigger_if = {
+					limit = {
+						exists = scope:has_selected_a_government
+					}
+					scope:has_selected_a_government = flag:japan_administrative
+				}
+				trigger_else_if = {
+					limit = {
+						top_liege = this
+					}
+					has_tgp_dlc_trigger = yes
+					tgp_can_become_japan_administrative_trigger = yes
+				}
+				trigger_else = {
+					scope:governor_liege = { government_has_flag = government_is_japan_administrative }
+				}
+			}
+			change_to_administrative_holding_effect = yes
+			tgp_domicile_conversion_when_changing_government_type = {
+				NEW_DOMICILE_TYPE = japanese_manor
+				NEW_GOVERNMENT_TYPE = japan_administrative_government
+			}
+			change_government = japan_administrative_government
+		}
+		else_if = { # Celestial
+			limit = {
+				trigger_if = {
+					limit = {
+						exists = scope:has_selected_a_government
+					}
+					scope:has_selected_a_government = flag:celestial
+				}
+				trigger_else_if = {
+					limit = {
+						scope:governor_liege ?= this
+					}
+					has_tgp_dlc_trigger = yes
+					OR = {
+						tgp_can_become_celestial_trigger = yes
+						#This is Mongol breakup and they are hegemon
+						AND = {
+							exists = scope:great_yuan_ruler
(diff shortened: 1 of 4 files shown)


common/scripted_triggers/unop_triggers.txt
```diff
@@ -59,3 +59,16 @@ unop_war_of_defiance_notified_player_trigger = {
 		}
 	}
 }
+
+# An independent nomad settling into administration should reach the Meritocratic Khanate (steppe_admin)
+# government - the natural steppe-admin form - rather than skipping it for plain Meritocratic/Administrative.
+# Used in change_to_administrative_effect to give the steppe_admin branch an independent-ruler resolution path,
+# mirroring the meritocratic/celestial/ritsuryo branches. Restricted to current nomads so other callers are unaffected.
```
pharaox:
```suggestion
# Used in change_to_administrative_effect to give the steppe_admin branch an independent-ruler resolution path,
# mirroring the other branches. Restricted to current nomads so other callers are unaffected.
```

pharaox (commented):
**reviewer:** Needs work — the approach is right and the mechanism is confirmed reachable, but the shared effect's blast radius isn't fully contained.

The core insight holds up: the steppe_admin branch really was the only one of its siblings without an independent-ruler resolution path (its `trigger_if` is hard `always = no` and its `trigger_else` needs an already-steppe_admin liege), and the new block mirrors the Celestial/Meritocratic template exactly, ordered ahead of Meritocratic. I traced the path end-to-end: `.0030` option b calls the effect with `flag:dynamic`, which explicitly *clears* `has_selected_a_government`, so the new `trigger_else_if` is genuinely evaluated.

- **Medium — Great Yuan breakup is plausibly re-routed.** `convert_great_yuan_to_admin_effect` calls `change_to_administrative_effect` on a ruler who is still nomadic, independent and TGP — matching the new trigger. The Celestial branch does *not* intercept, because its Great Yuan case requires `primary_title = title:h_china` while this path asserts `primary_title = title:e_great_yuan`. Previously they fell through to Meritocratic via `exists = scope:great_yuan_ruler`; now the new branch wins whenever the capital is still in `world_steppe`, which is likely since nothing relocates it beforehand and `great_yuan_de_jure_region` is Chinese, not steppe. That outcome may even be desirable, but it's untested and outside this issue's scope — worth either excluding explicitly or confirming in-game.

Everything else checked out: the gate is stricter than vanilla's own `adopt_steppe_administration` route (which has no region gate at all); the nested `every_vassal` callers are all vassals and can't match; the bundled `e_great_yuan` fix is an unambiguous typo (all eight siblings in that block use a direct `title:` reference); the `treasury = 0` suppression is a real false positive, confirmed by vanilla's own adjacent comment; `make tiger` re-run clean; and the vanilla file is byte-identical and in its own commit.

common/scripted_effects/00_administrative_effects.txt
```diff
@@ -0,0 +1,556 @@
...
+		else_if = { # Meritocratic Khanate
+			limit = {
+				trigger_if = {
+					limit = {
+						exists = scope:has_selected_a_government
+					}
+					# There's a separate decision to go from Nomad to Meritocratic Khanate, so you cannot pick it from here
+					always = no
+				}
+				trigger_else_if = { #Unop An independent settling nomad should reach the Meritocratic Khanate, not skip it for plain Meritocratic
```
pharaox, line 198: **reviewer:** *Medium.* This correctly targets the GOK reform, but `change_to_administrative_effect` has many callers and one of them matches this trigger unintentionally.

`convert_great_yuan_to_admin_effect` (`00_mongol_invasion_effects.txt:11878`, called from `:11368`) invokes this effect on the Great Yuan ruler while they are still nomadic, independent and TGP. The Celestial branch above does not catch them — its Great Yuan case requires `primary_title = title:h_china`, but this path asserts `primary_title = title:e_great_yuan` (`:11328`). So they now land here instead of Meritocratic (which previously caught them via `exists = scope:great_yuan_ruler`), provided their capital is still in `world_steppe` — likely, since no `set_realm_capital` runs before the conversion.

If the Great Yuan should keep its existing outcome, mirroring how the Meritocratic trigger special-cases that scope would contain it:

```suggestion
				trigger_else_if = { #Unop An independent settling nomad should reach the Meritocratic Khanate, not skip it for plain Meritocratic
					limit = {
						scope:governor_liege ?= this
						NOT = { exists = scope:great_yuan_ruler }
					}
					unop_should_become_steppe_admin_trigger = yes
				}
```

Otherwise it'd be worth confirming in-game that Great Yuan → Meritocratic Khanate is an acceptable (or intended) change.

common/scripted_triggers/unop_triggers.txt
```diff
@@ -59,3 +59,14 @@ unop_war_of_defiance_notified_player_trigger = {
...
+
+# Used in change_to_administrative_effect to give the steppe_admin branch an independent-ruler resolution path,
+# mirroring the other branches. Restricted to current nomads so other callers are unaffected.
+unop_should_become_steppe_admin_trigger = {
+	government_has_flag = government_is_nomadic
+	has_tgp_dlc_trigger = yes
+	is_independent_ruler = yes
+	capital_province = {
+		geographical_region = world_steppe
+	}
```
pharaox, line 71: **reviewer:** *Low.* This uses `world_steppe` (west/central/east/tarim) while the immediately-following Meritocratic branch's trigger uses only `world_steppe_east`.

The wider region is defensible — vanilla's own route to this government, `adopt_steppe_administration`, has no geographic gate at all, so this is still stricter than the established path. Flagging only so the asymmetry with the very next branch is a deliberate choice rather than an incidental one.

pharaox:
**fixer:** Addressed both review findings.

**Medium (Great Yuan re-routing) — fixed as suggested.** Verified the finding against the code before applying it, and it holds:

- `tgp_should_become_meritocratic_trigger` (`10_tgp_triggers.txt:526`) has an explicit third arm `exists = scope:great_yuan_ruler`, commented *"Becoming admin with creation of meritocratic great yuan"* — vanilla **deliberately** sends the Great Yuan breakup to Meritocratic.
- The Celestial branch above does not catch them: its Great Yuan case requires `primary_title = title:h_china`, but `convert_great_yuan_to_admin_effect` is reached under `primary_title = title:e_great_yuan` (`00_mongol_invasion_effects.txt:11328`).

So my branch would have silently overridden an intentional vanilla routing. Added `NOT = { exists = scope:great_yuan_ruler }` to the `limit`, mirroring how the Meritocratic trigger special-cases the same scope. The Great Yuan keeps its existing outcome.

**Low (region asymmetry) — deliberate, keeping `world_steppe`.** Vanilla's own route to this government, `adopt_steppe_administration`, has **no geographic gate at all** (`tgp_steppe_decisions.txt:19-27`: nomadic + kingdom/empire tier + culture head + nomadic authority + realm size), so the gate is already strictly narrower than the established path. The neighbouring `world_steppe_east` belongs to a broader China-adjacent set (`world_asia`, `world_tibet`, `world_burma`) aimed at rulers near Chinese meritocracy; Meritocratic Khanate is the steppe-native government, so a Mongol or Cuman GOK in the western/central steppe should reach it too. Narrowing to `world_steppe_east` would leave exactly those rulers falling through to plain Meritocratic — the bug this fixes.

`make tiger`: `fatal: 0, error: 0, warning: 0`. Still worth an in-game check with a settling nomad GOK, and with a Great Yuan breakup to confirm it still lands on Meritocratic.

common/scripted_effects/00_administrative_effects.txt
pharaox: **fixer:** Applied. Confirmed the finding first: `tgp_should_become_meritocratic_trigger` has an explicit `exists = scope:great_yuan_ruler` arm ("Becoming admin with creation of meritocratic great yuan"), so routing the Great Yuan to Meritocratic is intentional vanilla behaviour, and the Celestial branch does miss them because this path runs under `title:e_great_yuan`, not `title:h_china`. Added the `NOT = { exists = scope:great_yuan_ruler }` guard so their outcome is unchanged.

common/scripted_triggers/unop_triggers.txt
pharaox: **fixer:** Deliberate — keeping `world_steppe`. Vanilla's own route to this government (`adopt_steppe_administration`) has no geographic gate at all, so this is already stricter than the established path. The neighbouring `world_steppe_east` sits in a China-adjacent set (`world_asia`/`world_tibet`/`world_burma`) aimed at rulers bordering Chinese meritocracy, whereas Meritocratic Khanate is the steppe-native government — narrowing to `world_steppe_east` would leave a western/central-steppe Mongol or Cuman GOK falling through to plain Meritocratic, which is the bug being fixed.

pharaox (commented):
**reviewer:** Looks good — Fix 2 resolves the Medium precisely, and the Low was answered with a deliberate, verified decision.

Verified:
- **The guard fires on exactly the intended path.** `convert_great_yuan_to_admin_effect` is called at `00_mongol_invasion_effects.txt:11369`, nested inside `if = { limit = { exists = scope:great_yuan_ruler } ; scope:great_yuan_ruler = { ... } }` (11321–11325), so the new `NOT` reliably excludes the Great Yuan breakup.
- **And on no other path.** `scope:great_yuan_ruler` is saved in exactly one place in all of vanilla — `grant_great_yuan_effect` (`:3897`). The GOK reform never sets it, so the guard can't suppress the fix's own target case.
- **Fallthrough is correct.** With the arm's `limit` failing, a Great Yuan ruler drops to the branch's `trigger_else`, which needs `scope:governor_liege = { government_has_flag = government_is_steppe_admin }`. For an independent ruler `governor_liege` is `top_liege` = self, still nomadic (`00_administrative_effects.txt:58-60`), so it can't match — they fall through to Meritocratic exactly as before. Outcome genuinely unchanged.
- **Vanilla intent confirmed:** `tgp_should_become_meritocratic_trigger`'s `exists = scope:great_yuan_ruler` arm is commented "Becoming admin with creation of meritocratic great yuan", so that routing is deliberate and is now preserved.
- **Low accepted.** `world_steppe_east` sits alongside `world_asia`/`world_tibet`/`world_burma` in the meritocratic trigger — a China-adjacent set — while Meritocratic Khanate is the steppe-native government; narrowing would leave a western/central-steppe GOK falling through to plain Meritocratic, i.e. the reported bug. `adopt_steppe_administration` has no geographic gate at all, so the trigger stays strictly narrower than vanilla's own route.
- **Blast radius contained:** only `nomad_government` carries `government_is_nomadic`, plus TGP + independence + steppe capital.
- **Bundled Mongol fix completes its pattern class:** `main` already had two `#Unop` fixes for the same unset-`scope:great_yuan_title` bug (5597, 10592); this is the third and last. All remaining `scope:great_yuan_title` uses are inside `grant_great_yuan_effect`, where it is set.
- `make tiger` re-run independently: 0 errors, 0 warnings. The `treasury = 0` suppression is a genuine false positive (vanilla's own comment: "exactly 0 treasury, i.e. newly get this feature"). Vanilla file in `4005de9c` byte-identical to 1.19, added unmodified in its own commit.

**In-game test still worth doing:** confirm a settling nomad GOK lands in Meritocratic Khanate and a Great Yuan breakup still becomes Meritocratic. Resolution order is verified in script, but the resulting government is worth eyeballing once.

PR #607, opened 2026-07-25
Fix Roman restoration converting reformed Hellenists to the unreformed faith

### LINKED ISSUE #602: ep3_roman_restoration.0100 always sets faith to unreformed Hellenism even if reformed Hellenism exists
**Reported on Steam by `White Lich` on `2026-07-21 15:27`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> A surprising amount of content for Hellenism is bared behind being unreformed, palace_of_ctesiphon_02 is locked behind unreformed Hellenism. ep3_roman_restoration.0100 will always set your faith to unreformed Hellenism when you go for a true Roman Empire even if reformed Hellenism exists. Those are the two most notable, there are more, anything that is faith = faith:hellenic_pagan instead of faith = faith:hellenic_pagan.religion

pharaox:
Fixes #602

**fixer:** `ep3_roman_restoration.0100` options A ("LARP / Hard Mode") and C ("Roleplay") hardcoded `faith:hellenic_pagan` when setting character, state and county faith. That faith is specifically the *unreformed* one — it's the sole scripted faith in `hellenism_religion` and carries `unreformed_faith_doctrine`. A reformed Hellenic faith is a different faith object, so restoring Rome stripped reformed Hellenists of their own faith, permanently (the decision is once-per-game and irreversible).

The decision itself already admits reformed Hellenists at *religion* level — `restore_roman_empire_decision_religion_culture_trigger` tests `this = religion:hellenism_religion` — so the eligibility gate and the effect disagreed about what "Hellenic" means, and the effect was wrong.

**Fix:** skip the conversion when the ruler is already Hellenic, and derive the downstream faiths from the ruler instead of the literal:

- `set_character_faith` is now wrapped in `NOT = { faith.religion = faith:hellenic_pagan.religion }` — a reformed Hellenist keeps their faith; everyone else converts exactly as before.
- `set_state_faith` / `set_county_faith` now use `root.faith`, which resolves to the same `faith:hellenic_pagan` for every previously-working case, and to the retained reformed faith otherwise.
- The `doctrine_monotheist` check tested the literal `faith:hellenic_pagan` for a doctrine it can never have; it now tests the ruler's actual resulting faith.
- Vassal conversion needed no change — `ask_for_conversion_interaction` converts toward `scope:actor.faith`, so it follows the retained faith automatically.

Also fixed `ep3_roman_restoration.0101`'s desc trigger (same root cause, cosmetic): a reformed-Hellenic emperor fell through to the generic `fallback` text instead of the `hellenic` one.

This keeps to "fix, don't add" — it avoids inventing a policy for *which* reformed faith to pick when several exist, since no conversion is needed for someone already Hellenic.

Vanilla file added unmodified in a separate first commit. `make tiger` is clean.

Worth verifying in-game: with a reformed Hellenic faith, confirm options A/C preserve it and that state/county faith and vassal-conversion offers follow it; confirm unreformed Hellenic and Abrahamic rulers are unchanged.

### DIFF
--- a/events/dlc/ep3/ep3_roman_restoration_events.txt
+++ b/events/dlc/ep3/ep3_roman_restoration_events.txt
@@ -1,0 +1,533 @@
+﻿namespace = ep3_roman_restoration
+
+#######################################
+###
+###	EP3 RESTORE ROME STORY CYCLE EVENTS
+### by Chad Uhl
+###
+### 0100 		Justinian's Dream
+### 0101-0109 	Rome Reformed Notifications
+### 0300-0590 	Post-Restoration Difficulties
+###
+#######################################
+
+# Justinian's Dream
+# Event fired on story cycle setup informing the player that they can form the roman empire
+ep3_roman_restoration.0001 = {
+	type = character_event
+	title = ep3_roman_restoration.0001.t
+	desc = ep3_roman_restoration.0001.desc
+	theme = emperor
+
+	override_background = {
+		trigger = { capital_county = title:c_byzantion }
+		reference = ep3_constantinople
+	}
+
+	left_portrait = {
+		character = root
+		animation = emotion_thinking_scepter
+	}
+
+	trigger = {
+		NOT = { has_global_variable = flag_restored_roman_empire }
+	}
+
+	option = { 
+		name = ep3_roman_restoration.0001.a
+		custom_tooltip = ep3_roman_restoration.0001.a.tt
+		decision:restore_roman_empire_decision = {
+			open_view_data = {
+				view = decision_detail
+				player = root
+			}
+		}
+	}
+}
+
+# Should I LARP or nah?
+ep3_roman_restoration.0100 = {
+	type = character_event
+	window = fullscreen_event
+	title = ep3_roman_restoration.0100.t
+	desc = ep3_roman_restoration.0100.desc
+	theme = emperor
+	override_background = { reference = ep3_fullscreen_restore_rome }
+
+	immediate = {
+		save_scope_as = scoped_emperor
+	}
+
+	option = { # LARP (Hard Mode)
+		name = ep3_roman_restoration.0100.a
+		custom_tooltip = ep3_roman_restoration.hard_mode
+		custom_tooltip = ep3_roman_restoration.end_hard_mode
+		custom_tooltip = ep3_roman_restoration.invasion_cb
+		every_held_title = {
+			limit = {
+				is_head_of_faith = yes
+			}
+			root = {
+				destroy_title = prev
+			}
+		}
+		set_character_faith = faith:hellenic_pagan
+		if = {
+			limit = { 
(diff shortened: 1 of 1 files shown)


pharaox (commented):
**reviewer:** Looks good — the right hypothesis, and the whole pattern class within this feature is covered.

Skipping the conversion when the ruler is already Hellenic (religion-level) and deriving the downstream faiths from `root.faith` fixes the gate/effect disagreement without inventing a faith-selection policy.

Verified:
- **Root cause and intent.** `restore_roman_empire_decision_religion_culture_trigger` admits Hellenists at *religion* level, while options A/C converted at *faith* level — the gate and effect genuinely disagreed and the effect was wrong. The decision is once-per-game and irreversible, so the loss was permanent.
- **`root.faith` threading is idiomatic and correctly ordered.** The conditional conversion runs before the reads, so `root.faith` is the post-conversion faith. Both forms are established vanilla idioms — `ep3_decisions.txt:1781` for `set_state_faith = root.faith`, and `religion_on_actions.txt:104` uses `set_county_faith = root.faith` inside a county iterator exactly as here, confirming `root` isn't rebound by the iterator. Previously-working cases are unchanged since `root.faith` resolves to `faith:hellenic_pagan` for them.
- **Vassal conversion needed no change, as claimed.** The interaction runs with `actor = root`, and `new_faith_created_conversion_effect` converts via `set_county_faith = scope:actor.faith` — so offers follow the retained faith automatically.
- **The `doctrine_monotheist` change is a real improvement, not a behaviour change.** `doctrine_theism` is a `main_group` doctrine and Hellenism sets `doctrine_polytheist` at religion level with no faith-level override, so the old check was dead code. Testing `faith = { ... }` is the right level (Zoroastrianism/Dualism do declare theism per-faith) and still evaluates false for every Hellenic faith.
- **Pattern class complete.** Vanilla's EP3 file had 9 faith-pinned Hellenic sites; all 9 are addressed. Option B correctly untouched. The non-EP3 chain already uses `.religion` at 22 sites with zero faith-level ones. The two remaining literals (79, 147) are the intended conversion *target* for non-Hellenic rulers.
- **On the `destroy_title` block flagged as out-of-scope — there's actually no residual issue.** `doctrine_head_of_faith` is a `main_group` doctrine and Hellenism sets `doctrine_no_head` with no faith-level override, so a reformed Hellenic faith can't have a head-of-faith title. The block is a no-op for a reformed Hellenist and only fires for a converting Abrahamic ruler, which is intended.
- Vanilla file in `c56b59be` is **byte-exact** to 1.19, added unmodified in its own commit; CRLF+BOM preserved with no line-ending churn. `make tiger` re-run independently: 0 errors, 0 warnings.

**Worth confirming in-game:** with a reformed Hellenic faith, take options A and C — the faith is retained, state/county faith follow it, vassal offers target it. And that an unreformed Hellenist and an Orthodox actor behave exactly as before (including head-of-faith title destruction for the latter).

PR #608, opened 2026-07-25
Fix terrain inspector decisions being available to all landless governments

### LINKED ISSUE #600: Unlanded Admin Head of Families can take choose_designated_terrain_decision
**Reported on Steam by `White Lich` on `2026-07-22 02:46`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> Unlanded Admin Head of Families can take choose_designated_terrain_decision despite being intended to be locked to Adventurers.

pharaox:
Fixes #600

**fixer:** `choose_designated_terrain_decision` gated on `is_landed = no`, which is much wider than "adventurer". Six governments set `landless_playable = yes` — including `administrative_government` — so an unlanded Administrative Head of Family passes the gate while being neither adventurer nor nomad. Nothing downstream re-checks: `wanderer_lifestyle` is gated only on `has_dlc_feature = wandering_nobles`, so they can genuinely take `mustering_the_troops_perk` and reach the decision.

**The effect layer shows the intended audience unambiguously.** In `landless_inspector_assign_terrain_effect`, every tier-2 and tier-3 branch requires `government_is_landless_adventurer` or `government_is_nomadic`; only the final `else_if` has no government check, and that is what an Admin HoF falls through to. In `landless_inspector_upgrade_terrain_tier_extrapolated_effect`, *all four* upgrade branches require adventurer or nomad. So an affected character gets a tier-1 modifier and is then permanently stuck there — the `local_inspection_perk` and `no_stone_unturned_perk` payoffs can never apply. The decision is also explicitly a permanent choice, so that dead end is irreversible.

**Fix:** replaced `is_landed = no` with vanilla's own idiom for this, `is_landless_adventurer = yes`, which `enact_legacy_of_adventurer_decision` uses in the same file. The `government_is_nomadic` branch is preserved — nomads are landed, which is the only reason the `OR` existed.

Also fixed `choose_master_terrain_decision` (same root cause, found during analysis): it gated on a bare `is_landed = no` with no nomad branch at all, so it both leaked to the same governments and would have needed the nomad branch anyway. It's reachable only by someone holding a tier-1+ inspector modifier — precisely the population the first bug creates — so fixing only the first decision would leave this one reachable in existing saves. Both now use the same explicit adventurer-or-nomad test.

Vanilla file added unmodified in a separate first commit. `make tiger` is clean.

Note: a character who already took the decision keeps a tier-1 modifier; cleaning up existing saves is out of scope for a guard fix.

Worth verifying in-game: as an unlanded Admin Head of Family with `mustering_the_troops_perk`, confirm the decision no longer appears; as a landless adventurer and as a nomad, confirm both decisions still do.

### DIFF
--- a/common/decisions/dlc_decisions/bp3/00_bp3_other_decisions.txt
+++ b/common/decisions/dlc_decisions/bp3/00_bp3_other_decisions.txt
@@ -1,0 +1,994 @@
+﻿### End Pious Investments ###
+end_investments_minor_decision = {
+	picture = {
+		reference = "gfx/interface/illustrations/decisions/decision_spend_money.dds"
+	}
+	sort_order = 1
+
+	is_shown = {
+		has_character_modifier = church_investment_cost_modifier
+	}
+
+	effect = {
+		remove_character_modifier = church_investment_modifier
+		remove_character_modifier = church_investment_cost_modifier
+	}
+
+	ai_potential = {
+		always = no
+	}
+
+	ai_check_interval = 0
+}
+
+### Designated Terrain ###
+choose_designated_terrain_decision = {
+
+	ai_check_interval_by_tier = {
+		barony = 0
+		county = 0
+		duchy = 120
+		kingdom = 0
+		empire = 0
+		hegemony = 0
+	}
+
+	picture = { reference = "gfx/interface/illustrations/decisions/ep3_decision_laamp_neutral_mountain.dds" }
+
+	sort_order = 10
+	decision_group_type = major
+
+	is_shown = {
+		OR = {
+			is_landed = no
+			government_has_flag = government_is_nomadic
+		}
+		has_perk = mustering_the_troops_perk
+		NOR = {
+			landless_inspector_terrain_master_trigger = { TERRAIN = woodlands }
+			landless_inspector_terrain_master_trigger = { TERRAIN = highlands }
+			landless_inspector_terrain_master_trigger = { TERRAIN = lowlands }
+			landless_inspector_terrain_master_trigger = { TERRAIN = drylands }
+		}
+	}
+
+	widget = {
+		gui = "decision_view_widget_option_list_generic"
+		controller = decision_option_list_controller
+		decision_to_second_step_button = "CHOOSE_DESIGNATED_TERRAIN_DECISION_NEXT_STEP_BUTTON"
+		show_from_start = yes
+
+		item = { # Woodlands
+			value = master_forest_terrain
+			current_description = designated_terrain_forest_decision
+			localization = designated_terrain_forest_decision
+			icon = "gfx/interface/icons/terrain_types/forest.dds"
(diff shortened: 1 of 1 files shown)


pharaox (commented):
**reviewer:** Looks good — correct root cause and altitude, uses vanilla's own idiom from the same file, and both members of the pattern class are fixed.

The decisive evidence is the government table. Enumerating the definitions programmatically, **six** governments set `landless_playable = yes` — `administrative`, `landless_adventurer`, `nomad`, `celestial`, `steppe_admin`, `meritocratic` — but only `landless_adventurer_government` has `government_is_landless_adventurer` and only `nomad_government` has `government_is_nomadic`. So the narrowing admits exactly the two governments the effect layer pays out to and excludes exactly the four that got nothing but a permanent dead end.

Verified:
- **The effect layer unanimously establishes intent.** In `landless_inspector_assign_terrain_effect` all four tier-2/3 branches require adventurer or nomad; only the terminal tier-1 `else_if` lacks a government check. In `landless_inspector_upgrade_terrain_tier_extrapolated_effect` **all four** upgrade branches require adventurer or nomad — so a leaked-in ruler is permanently stuck at tier 1, on an irreversible choice. Two guards vs. the whole effect layer; the fix follows the effect layer.
- **`is_shown` is the only control point** — neither decision has an `is_valid`, and `wanderer_lifestyle` is gated solely on the DLC feature, so nothing downstream would have caught this.
- **The added `OR` on `choose_master_terrain_decision` is necessary, not incidental.** It had a *bare* `is_landed = no` with no nomad branch, so narrowing to `is_landless_adventurer = yes` alone would have locked nomads out. Independently corroborated that nomads belong there: `landless_inspector_terrain_master_trigger`, which that decision requires, explicitly tests `nomadic_inspector_$TERRAIN$_tier_2/tier_3_modifier`. Good catch.
- **Fixing both in one PR is the right scope** — the master decision requires an existing tier-1+ modifier, so it's only reachable by someone who already took the first decision, i.e. exactly the population this bug creates.
- **Pattern class complete.** Exactly two `is_landed = no` sites exist in the vanilla file; both fixed. Grepping the rest of vanilla for other entry points into the feature turns up only unrelated uses (inspection activity, script values, perk effect loc).
- `is_landless_adventurer` is a one-line wrapper over the government flag, and `enact_legacy_of_adventurer_decision` in the same file (line 921) already uses it — vanilla's own vocabulary.
- Vanilla file in `200d4234` is **byte-exact** to 1.19, added unmodified in its own commit. `make tiger` re-run independently: 0 errors, 0 warnings.

Agreed on leaving the residual tier-1 modifier for characters who already took the decision — cleanup would read as an addition, and the modifier is harmless.

**Worth confirming in-game:** as an unlanded Administrative Head of Family with `mustering_the_troops_perk`, neither decision appears; as a landless adventurer and as a nomad, both still do.

PR #622, opened 2026-08-01
Fix physician treatment events not firing when appointed outside the hire decision

### LINKED ISSUE #605: health.3101 physician treatment event does not fire unless the physician was recruited for a fee
**Reported on Steam by `=Sarsante=` on `2026-07-23 17:57`** ([source](https://steamcommunity.com/workshop/filedetails/discussion/2871648329/4697908845434539387/)).

> Also this one that I posted here a long time ago and there's no reply, maybe you guys missed it.
>
> I don't know if this is a bug, an overview or if it's done for performance reasons but the event health.3101, event that your physician will treat your character don't fire unless you use the recruit physician and pay for it.
>
> If you've disease and appoint someone as physician or use the find physician option but invite them to court instead of paying the fee the event will not fire ever.
>
> A good way of testing it is start a Baldwin run in 1178 then you can use the decision and invite someone or use character finder and move them to your court, when they're employed the event to get treated from lepracy will not fire ever.
>
> If you follow the "normal" path, look for physician and pay their fee the event will fire right away.
>
> In debug it's possible to see the conditions for the event to fire which are basically employ physician and have one of many diseases but that's not what happens in-game.
>
> Could you please check it out? Probably a monthly/yearly check it's broken.
>
> I can provide more explaination if needed. Thanks.

pharaox:
Fixes #605

**fixer:** `health.3101`/`health.4001` treatment events only fired when the court physician was hired through the paid `health.3001` decision chain. The kick-off logic lived inside `set_court_physician_effect`, which is only reachable from that chain's event options — appointing an existing courtier via the court-position UI went through `on_court_position_received` instead, which was a bare pass-through. Since the `health.3100` gate self-terminates when no physician is available, the player was then stuck permanently.

Moved the treatment kick-off to the position's accepted hook, which the engine runs for *every* appointment route:

- Extracted the block verbatim into `unop_trigger_physician_treatments_effect` in `00_unop_effects.txt`.
- Commented out the inline copy in `set_court_physician_effect` and call the new effect from `court_physician_title_accepted_effect` on `scope:liege` instead.

Moving rather than duplicating avoids double-firing on the paid path: `set_court_physician_effect` itself calls `appoint_court_position`/`replace_court_position`, so the hook already covers it, and `has_recent_treatment_trigger` only checks post-treatment modifiers — it would not have suppressed a second queued event.

This makes the vanilla comment at `health_events.txt:7300` ("if we hire a new physician, they will treat everyone upon being hired") actually true, and mirrors how `wet_nurse_title_accepted_effect` already puts its on-hire logic in the hook.

### DIFF
--- a/common/scripted_effects/00_unop_effects.txt
+++ b/common/scripted_effects/00_unop_effects.txt
@@ -739,3 +739,50 @@ unop_distribute_determined_court_position_effect = {
 		}
 	}
 }
+
+#Unop Trigger physician treatments for the employer and their court; called from the court physician's on_court_position_received hook so every appointment route works, not just the paid hire event
+unop_trigger_physician_treatments_effect = {
+	hidden_effect = {
+		if = {
+			limit = {
+				has_treatable_disease_trigger = yes
+				has_recent_treatment_trigger = no
+			}
+			save_scope_as = sick_character
+			trigger_event = {
+				id = health.3101
+				days = { physician_treatment_delay_ruler_min physician_treatment_delay_ruler_max }
+			}
+		}
+		else_if = {
+			limit = {
+				has_recent_wound_treatment_trigger = no
+				has_wounds_trigger = yes
+			}
+			save_scope_as = sick_character
+			trigger_event = {
+				id = health.4001
+				days = { physician_treatment_delay_ruler_min physician_treatment_delay_ruler_max }
+			}
+		}
+		every_courtier_or_guest = {
+			if = {
+				limit = {
+					has_treatable_disease_trigger = yes
+					has_recent_treatment_trigger = no
+				}
+				set_worst_disease_effect = yes
+				save_scope_as = sick_character
+				decide_who_picks_disease_treatment_effect = yes #Sends health.3101 to them or 3102 to liege (you)
+			}
+			else_if = {
+				limit = {
+					has_recent_wound_treatment_trigger = no
+					has_wounds_trigger = yes
+				}
+				save_scope_as = sick_character
+				decide_who_picks_wound_treatment_effect = yes
+			}
+		}
+	}
+}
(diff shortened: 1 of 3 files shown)


common/scripted_effects/00_unop_effects.txt
```diff
@@ -738,4 +738,51 @@ unop_distribute_determined_court_position_effect = {
...
+				limit = {
+					has_recent_wound_treatment_trigger = no
+					has_wounds_trigger = yes
+				}
+				save_scope_as = sick_character
+				decide_who_picks_wound_treatment_effect = yes
+			}
+		}
+	}
 }
```
pharaox, line 788: **reviewer:** `Low` — the file still ends without a trailing newline (`\ No newline at end of file`). It was already missing before this PR, so this is inherited rather than introduced, but since the new effect is appended right here it is a cheap tidy-up.

pharaox (commented):
**reviewer:** Looks good — the fix targets the root cause at the right altitude and I could not break it.

Verified beyond the diff:

- `on_court_position_received` (`00_court_positions.txt:432`) is the **only** call site of `court_physician_title_accepted_effect`, and there is no Celestial/Mandala/Admin physician variant — the pattern class is fully enumerated.
- The load-bearing assumption (the hook fires for `appoint_court_position`/`replace_court_position`) is corroborated by `_court_positions.info:254-267` and by the `wet_nurse_title_accepted_effect` precedent, which uses the identical `scope:liege = { ... }` shape.
- **Paid path is unchanged.** All six callers of `set_court_physician_effect` reach an appoint/replace, so they still reach the block via the hook. Moving rather than mirroring correctly avoids the double-fire risk flagged during analysis.
- **No spurious firing.** `health.3101` re-validates `court_physician_available_trigger` in its own `trigger` and re-resolves the physician in `immediate`, so an appointment undone within the 2-4 day delay is filtered by the event itself.

One `Low` note inline; it does not affect correctness.

PR #623, opened 2026-08-01
Fix retirement decisions not transferring gold to the heir

### LINKED ISSUE #609: Retire from heading the house decision doesn't transfer gold to the new house head
**Reported on Steam by `jy` on `2026-07-26 02:37`** (<https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329>).

> the retire from heading the house decision doesnt pass the cash over, new character starts with 0 cash

pharaox:
Fixes #609

**fixer:** `tgp_renounce_estate_effect` moves held titles, vassals, house headship and player control to the heir, but never the retiree's wealth — so the new player character starts broke while the retiree lives on as a `devoted` courtier holding the whole treasury. The decision is a scripted imitation of succession, so anything the engine does implicitly on death has to be done explicitly here, and gold was overlooked.

Added a transfer next to `set_house_head`, inside the existing `exists = scope:new_player` branch:

- Uses `pay_treasury_or_gold` because these are administrative-family rulers (`decision_group_type = admin`); Administrative/Celestial/Meritocratic/Steppe governments spend *treasury*, so a bare `add_gold` would mishandle them.
- The value branches on `has_treasury`, mirroring the vanilla `*_treasury_or_gold_value` idiom.
- Guarded with a `> 0` check, matching `07_ep3_wars.txt:5119` — without it a retiree in debt would drain the heir instead.

Full transfer matches the decision's loc ("Entrust everything to your house's heir"); a partial remainder would be a balance choice and out of scope. Fixing the shared effect covers both entry points — the retirement decision and `tgp_japan_become_a_monk_decision`.

Closest vanilla precedent is `game_start.txt:4592`, which handles the same title-handover-plus-player-swap and pairs it with `pay_short_term_gold = { gold = gold }`.

### DIFF
--- a/common/scripted_effects/10_dlc_tgp_japan_scripted_effects.txt
+++ b/common/scripted_effects/10_dlc_tgp_japan_scripted_effects.txt
@@ -3910,6 +3910,31 @@ tgp_renounce_estate_effect = {
 				set_house_head = scope:new_player
 			}
 		}
+		#Unop Also hand over the treasury/gold, otherwise the heir inherits everything except the wealth
+		if = {
+			limit = {
+				trigger_if = {
+					limit = { has_treasury = yes }
+					treasury > 0
+				}
+				trigger_else = {
+					gold > 0
+				}
+			}
+			pay_treasury_or_gold = {
+				target = scope:new_player
+				value = {
+					value = 0
+					if = {
+						limit = { has_treasury = yes }
+						add = treasury
+					}
+					else = {
+						add = gold
+					}
+				}
+			}
+		}
 		if = {
 			limit = {
 				exists = scope:new_human_player_character


common/scripted_effects/10_dlc_tgp_japan_scripted_effects.txt
```diff
@@ -3910,6 +3910,31 @@ tgp_renounce_estate_effect = {
...
+			limit = {
+				trigger_if = {
+					limit = { has_treasury = yes }
+					treasury > 0
+				}
+				trigger_else = {
+					gold > 0
+				}
+			}
+			pay_treasury_or_gold = {
```
pharaox, line 3924: **reviewer:** Not a finding — a note for in-game testing. The decision’s `show_as_tooltip` preview in `tgp_dynastic_cycle_decisions.txt` was deliberately left unextended, on the reasoning that `pay_treasury_or_gold` emits its own tooltip line here. Agreed that is the minimal change; worth an eyeball in-game to confirm the gold handover actually reads clearly to the player before they commit to the decision.

pharaox (commented):
**reviewer:** Looks good — correct fix at the right altitude, and the load-bearing assumption checks out against the game docs.

Verified beyond the diff:

- **All entry points covered.** `tgp_renounce_estate_effect` has **three** vanilla call sites, not two: `tgp_japan_decisions.txt:1346`, `tgp_dynastic_cycle_decision_events.txt:23`, and `tgp_japan_decision_events.txt:2335`. Fixing the shared effect covers all three.
- **`pay_treasury_or_gold` semantics confirmed from `logs/effects.log`**, not just by precedent: *"the scoped character pays treasury if exists, otherwise gold; the target character receives treasury if exists, otherwise gold."* It debits the payer (wealth moves, not duplicated) **and** resolves payer/recipient currency independently — which matters here, because the heir may not have assumed a treasury government yet at transfer time. Nothing is lost either way.
- **No budget truncation.** I checked whether the `pay_*` family would cap a full-balance transfer at the AI "short term" budget; `effects.log` says these take from the category *"first, then rest"*, so the whole amount is paid.
- **Ordering/scope are right.** The block runs in the retiree scope after titles transfer and before `set_player_character`; `has_treasury` is a government property, not title-derived, so it still reads correctly post-transfer.
- The `> 0` guard is a real catch — without it a retiree in debt would drain the heir.

No findings. One note for in-game testing inline.

PR #625, opened 2026-08-01
Fix civil war authority loss skipping Administrative and Nomad rulers

### LINKED ISSUE #612: Faction/depose/refused-demand wars only lower crown authority for Feudal and Tribal, excluding Admin, Celestial, Meritocratic, and Japanese
**Reported on Steam by `White Lich` on `2026-07-27 10:10`** (<https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329>).

> independence_faction_war, depose_war, and refused_liege_demand_war only lower crown authority for Feudal and Tribal. Excluding Admin, Celestial, Meritocratic, and Japanese.

pharaox:
Fixes #612

**fixer:** `independence_faction_war` and `refused_liege_demand_war` each contained an identical inline six-`if` chain stepping `crown_authority` and `tribal_authority` only. Because `imperial_bureaucracy` and `nomadic_authority` are separate law groups, every `if` silently no-opped for those governments — an Administrative ruler could lose a civil war with no authority penalty at all.

Replaced both inline chains with the government-agnostic helpers, which is the idiom vanilla already uses elsewhere (`00_diarch_interactions.txt`, `hold_court_events_general.txt`):

```
decrease_crown_authority_effect = yes
decrease_tribal_authority_effect = yes
decrease_imperial_bureaucracy_effect = yes
decrease_nomadic_authority_effect = yes
```

Each helper self-gates on its own `has_realm_law_flag = uses_*`, and the four flags sit on disjoint government types, so exactly one applies and the rest no-op. This also fixes Nomad/Herder, which the report didn't mention.

**Scope notes:**
- **`depose_war` deliberately excluded.** It has *no* realm-law change at all (verified: zero `realm_law` hits in its block), so it never lowered authority for anyone — that's not the reported asymmetry, and adding a penalty would be a balance change rather than a fix.
- **Celestial / Meritocratic / Steppe-Admin / Japanese cannot be fixed here.** `realm_law_use_imperial_bureaucracy` explicitly excludes them, so they have no authority ladder to lower. Giving them a penalty would need a different currency — new content, out of scope. The issue title overstates what is fixable.

Also fixed a pre-existing vanilla bug surfaced by adding this file as an override: `character = this` inside `every_character_struggle` (l.811) passed the *struggle* where a character was expected. Changed to `scope:defender`, matching vanilla's own `character = prev` idiom in `county_on_actions.txt:26`. Tiger is clean.

### DIFF
--- a/common/casus_belli_types/00_civil_war.txt
+++ b/common/casus_belli_types/00_civil_war.txt
@@ -1,0 +1,2608 @@
+﻿# Triggers as a result of the Independence Faction demand
+# Note: Wars for Independence started by individual characters is a separate CB
+independence_faction_war = {
+	icon = independence_faction_war
+	group = civil_war
+	ai_only_against_liege = yes
+	target_titles = independence_domain
+	allow_hostages = no
+	allowed_for_character = {
+		scope:attacker = {
+			is_leading_faction_type = independence_faction
+			NOT = {
+				government_has_flag = government_is_landless_adventurer
+			}
+		}
+	}
+
+	allowed_against_character = {
+		scope:attacker = {
+			liege = scope:defender
+		}
+	}
+
+	target_de_jure_regions_above = yes
+
+	valid_to_start = {
+
+	}
+
+	on_declaration = {
+		#on_declared_war = yes
+		# Remove offending HumSac modifiers.
+		scope:attacker = {
+			hidden_effect = { fp1_remove_humsac_offended_counties_effect = yes }
+		}
+
+		# If no faction exists because we created it through history, we add a flag on the attackers to free them at the end
+		if = {
+			limit = { NOT = { exists = scope:attacker.joined_faction } }
+			war = {
+				every_war_attacker = {
+					if = {
+						limit = { NOT = { has_character_flag = should_become_independent } }
+						add_character_flag = should_become_independent
+					}
+				}
+			}
+		}
+	}
+
+	on_victory_desc = {
+		first_valid = {
+			triggered_desc = {
+				trigger = { # Desc for only one player attacker
+					scope:attacker = {
+						is_local_player = yes
+						exists = joined_faction
+						joined_faction = {
+							any_faction_member = {
+								count = 1
+							}
+						}
+					}
+				}
+				desc = independence_war_victory_desc_local_player_attacker_alone
+			}
+			triggered_desc = { # Desc for only one attacker, player defender
+				trigger = {
+					scope:defender = {
+						is_local_player = yes
+					}
+					scope:attacker = {
+						exists = joined_faction
+						joined_faction = {
+							any_faction_member = {
+								count = 1
+							}
+						}
+					}
+				}
+				desc = independence_war_victory_desc_local_player_defender_attacker_alone
+			}
+			triggered_desc = { # Desc for player attacker
+				trigger = {
+					OR = {
+						AND = {
+							exists = scope:attacker.joined_faction
+							scope:attacker.joined_faction = {
+								any_faction_member = { is_local_player = yes }
+							}
+						}
+						scope:attacker = {
+								is_local_player = yes
+								has_character_flag = should_become_independent
+							}
+					}
+				}
+				desc = independence_war_victory_desc_local_player_attacker
+			}
+			triggered_desc = { # Desc for player defender
+				trigger = { scope:defender = { is_local_player = yes } }
+				desc = independence_war_victory_desc_local_player_defender
+			}
+			desc = independence_war_victory_desc # Desc for a third party involved
+		}
+
+	}
+
+	on_victory = {
+		scope:attacker = { show_pow_release_message_effect = yes }
+
+		#EP2 Accolade glory gain from winning against higher ranked enemy
+		scope:attacker = { accolade_attacker_war_end_glory_gain_med_effect = yes }
+
+		create_title_and_vassal_change = {
+			type = independency
+			save_scope_as = change
+		}
+		if = {
+			limit = { exists = scope:attacker.joined_faction }
+			scope:attacker.joined_faction = {
+				every_faction_member = {
+					change_liege_or_become_independent = {
+						CHANGE = scope:change
+						VASSAL = this
+					}
+
+					hidden_effect = {
+						set_variable = {
+							name = independence_war_former_liege
+							value = scope:defender
+						}
+						add_truce_both_ways = {
+							character = scope:defender
+							days = 1825
+							war = root.war
+							result = victory
+						}
+						save_scope_as = current_member
+
+						# Struggle Catalyst
+						if = {
+							limit = {
+								catalyst_gave_independence_to_powerful_diff_faith_culture_vassal_preliminary_trigger = {
+									CHAR1 = scope:defender
+									CHAR2 = scope:current_member
+								}
+								any_character_struggle = {
+									involvement = involved
+									activate_struggle_catalyst_secondary_character_involvement_involved_trigger = {
+										CATALYST = catalyst_gave_independence_to_powerful_diff_faith_culture_vassal
+										CHAR = scope:current_member
+									}
+								}
+							}
+							every_character_struggle = {
+								involvement = involved
+								limit = {
+									activate_struggle_catalyst_secondary_character_involvement_either_trigger = {
+										CATALYST = catalyst_gave_independence_to_powerful_diff_faith_culture_vassal
+										CHAR = scope:current_member
+									}
+								}
+								activate_struggle_catalyst = {
+									catalyst = catalyst_gave_independence_to_powerful_diff_faith_culture_vassal
+									character = scope:defender
+								}
+							}
+						}
+					}
+				}
+			}
+		}
+		else_if = {
+			# Free character with the flag
+			limit = { exists = war }
+			war = {
+				every_war_attacker = {
+					if = {
+						limit = { has_character_flag = should_become_independent }
+
+						change_liege_or_become_independent = {
+							CHANGE = scope:change
+							VASSAL = this
+						}
+
+						hidden_effect = {
+							set_variable = {
+								name = independence_war_former_liege
+								value = scope:defender
+							}
+							add_truce_both_ways = {
+								character = scope:defender
+								days = 1825
+								war = root.war
+								result = victory
+							}
+							save_scope_as = current_member
+
+							# Struggle Catalyst
+							if = {
+								limit = {
+									catalyst_gave_independence_to_powerful_diff_faith_culture_vassal_preliminary_trigger = {
+										CHAR1 = scope:defender
+										CHAR2 = scope:current_member
+									}
+									any_character_struggle = {
+										involvement = involved
+										activate_struggle_catalyst_secondary_character_involvement_involved_trigger = {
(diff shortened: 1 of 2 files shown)


common/casus_belli_types/00_civil_war.txt
```diff
@@ -0,0 +1,2618 @@
...
+#				add_realm_law = tribal_authority_1
+#			}
+#			if = {
+#				limit = { has_realm_law = tribal_authority_3 }
+#				add_realm_law = tribal_authority_2
+#			}
+			decrease_crown_authority_effect = yes
+			decrease_tribal_authority_effect = yes
+			decrease_imperial_bureaucracy_effect = yes
+			decrease_nomadic_authority_effect = yes
```
pharaox, line 274: **reviewer:** `Medium` — `decrease_nomadic_authority_effect` (`00_realm_effects.txt:164-181`) has branches for `nomadic_authority_4/3/2` only. `nomadic_authority` runs **1-5** (`00_realm_laws.txt:1500-1791`), and tier 5 is reachable — `can_pass = { has_realm_law = nomadic_authority_4 ... }` at l.1824, flagged `max_authority_level`. A tier-5 nomad therefore matches no branch and loses no authority from an independence faction war or a refused liege demand.

Vanilla handles exactly this case in its own nomad CBs, `09_mpo_wars.txt:1533-1538`:

```
switch = {
   trigger = has_realm_law
   nomadic_authority_5 = { add_realm_law = nomadic_authority_4 }
   nomadic_authority_4 = { add_realm_law = nomadic_authority_3 }
   nomadic_authority_3 = { add_realm_law = nomadic_authority_2 }
}
```

Since this PR is the first caller of that helper anywhere, the cleanest fix is to close the gap in the helper itself — `00_realm_effects.txt` is already an Unop override, and its own l.120 comment says these indirections exist precisely so they can be overridden. Adding the missing top branch there also keeps every future caller correct:

```
decrease_nomadic_authority_effect = {
	if = {
		limit = { has_realm_law_flag = uses_nomadic_authority }
		#Unop Handle the top tier too; nomadic_authority runs 1-5
		if = {
			limit = { has_realm_law = nomadic_authority_5 }
			add_realm_law = nomadic_authority_4
		}
		else_if = {
			limit = { has_realm_law = nomadic_authority_4 }
			add_realm_law = nomadic_authority_3
		}
		...
	}
}
```

If tier 5 is meant to be exempt, that is worth an explicit note instead — but the `09_mpo_wars.txt` precedent suggests it is not.

pharaox (commented):
**reviewer:** Needs work — the approach is right, but the Nomad half of the fix does not actually land for the strongest nomads.

**Medium — `nomadic_authority_5` is not handled.** `nomadic_authority` has **five** tiers, but `decrease_nomadic_authority_effect` only steps down from tier 4. A nomad at `nomadic_authority_5` (the `max_authority_level`) falls through every branch and loses nothing — so the claim that this "fixes Nomad as a side effect" holds for tiers 2-4 but not for the rulers most likely to actually face these wars. Vanilla’s own nomad CBs handle it: `09_mpo_wars.txt:1533-1538` switches from `nomadic_authority_5` down. Worth stressing: **this PR is the first caller of `decrease_nomadic_authority_effect` anywhere in the game** (zero vanilla callers), so the helper’s gap has never been exercised and "it is vanilla’s helper" is not cover here. Details inline.

Everything else checks out:

- **Altitude is right** — self-gating helpers over inline literal ladders, matching `00_diarch_interactions.txt:2494-2503`.
- **Mutual exclusivity confirmed** via `00_law_triggers.txt:64-88`: the four gates rest on disjoint government flags, so no ruler can lose two ladders.
- **Behavioural equivalence verified** for crown/tribal — ascending plain-`if`s and descending `else_if`s both step exactly one tier from any start.
- **`depose_war` correctly excluded** (zero `realm_law` changes today, so adding one is a balance change, not a fix), and Celestial/Meritocratic/Steppe-Admin/Japanese correctly identified as having no ladder at all.
- **Vanilla commit `dc29e02d` is byte-identical** to the game file (verified with `diff`), fix isolated in `9dd4564c`.
- **The `character = this` fix at l.811 is correct** and rightly fixed rather than suppressed.

pharaox:
**fixer:** Re-fix addressing the Medium review finding.

**Nomads at maximum authority now lose a tier.** Added the missing `nomadic_authority_5` branch to `decrease_nomadic_authority_effect` in `common/scripted_effects/00_realm_effects.txt` (already an Unop override, and the file's own l.120 comment endorses overriding these helpers). The existing first branch became `else_if` so the chain stays mutually exclusive and still steps exactly one tier.

Verified the finding rather than taking it on trust:
- `nomadic_authority_5` exists (`00_realm_laws.txt:1791`), carries `flag = max_authority_level`, and is reachable via `can_pass = { has_realm_law = nomadic_authority_4 }`.
- The helper stopped at tier 4, so a tier-5 nomad fell through every branch and lost nothing.
- Vanilla's own nomad CB does handle it: `09_mpo_wars.txt:1533-1538` opens its `switch` with `nomadic_authority_5 = { add_realm_law = nomadic_authority_4 }`. The fix mirrors that.
- Confirmed **zero vanilla callers** of `decrease_nomadic_authority_effect`, so this PR remains its first caller and the added branch cannot regress existing behaviour.

**Scope note:** `increase_nomadic_authority_effect` has the mirror-image gap (it stops at tier 4, never reaching 5). I left it alone deliberately — nothing calls it on this path, so patching it would be an unverifiable addition rather than a fix. Worth a separate issue if wanted.

Also **rebased onto current `main`** (the branch had fallen behind). History is unchanged in shape: vanilla file added unmodified in `8b2ce467`, fixes in `38a71543` and `60f99093`.

**Tiger:** `fatal: 0, error: 0, warning: 0, untidy: 0, tips: 0`.

**In-game check still worth doing:** confirm a nomad ruler at `nomadic_authority_5` losing an independence faction war or refused-demand war drops to tier 4.

pharaox (commented):
**reviewer:** Looks good — Review 1's `Medium` is resolved. One `Low` on the PR prose only (see below); it doesn't affect the fix or the verdict.

**Review 1 `Medium` (nomads at max authority lose nothing) — resolved.** The helper now opens with a `nomadic_authority_5 -> 4` branch, with the former first branch demoted to `else_if` so the chain stays mutually exclusive and steps exactly one tier. Re-confirmed the gap was live: tier 5 exists, carries `max_authority_level`, and is reachable via `can_pass = { has_realm_law = nomadic_authority_4 }`. Vanilla's `09_mpo_wars.txt:1533-1538` switch opens with exactly this transition, which also corroborates that `add_realm_law` replaces rather than stacks.

Verified independently this cycle:

- **No remaining top-tier gap anywhere else.** Enumerated all four ladders: crown 0-3, tribal 0-3, imperial_bureaucracy 0-3 — each fully covered by its three-branch helper. `nomadic_authority` was the **only** five-tier ladder, so it was the only one with a gap. Floor at tier 1 is right (there is no `nomadic_authority_0`).
- **Zero regression risk.** `decrease_/increase_nomadic_authority_effect` have **no vanilla callers** — this PR is the first, so the new branch can't change existing behaviour. Leaving `increase_*` alone is the right call.
- **Mutual exclusivity re-derived.** `government_is_nomadic` sits on `nomad_government` only, whose flag block carries no crown/tribal/administrative flag; the four gates are genuinely disjoint, so at most one ladder is ever lowered.
- **No missed call sites.** The only other inline `crown_authority` write is `07_frankokratia_scripted_effects.txt:2368`, an unconditional *set* during setup — different mechanic, correctly out of scope. Within this file every remaining inline ladder line is commented out, so both CBs were fully converted.
- **`depose_war` exclusion re-confirmed** — zero `realm_law` hits in l.1221-1461, so it never lowered authority for anyone. Right call to flag rather than bundle.
- **Helper calls are in `scope:defender`**, the same character scope as the code they replace.
- **`character = this` -> `scope:defender` is correct** — inside `every_character_struggle`, `this` is the struggle. Vanilla's `county_on_actions.txt:22-27` uses the same construct with `character = prev`.
- **Vanilla commit `8b2ce467` is byte-identical** to the game file, with fixes isolated in later commits. Branch is current with `main`; tiger `No problems found`, no suppressions.

---

**`Low` — PR body scope note is inaccurate (prose only, no code change needed):** the body says this "also fixes Nomad/Herder". Herders aren't fixed and can't be — `herder_government` carries `government_is_herder`, **not** `government_is_nomadic` (`00_government_types.txt:674-717`), and `realm_law_use_nomadic_authority` gates solely on the latter. `government_is_herder` appears in exactly one trigger game-wide, `can_have_herder_succession_law_trigger`, which is a *succession* law. Herders have no authority ladder at all and belong with Celestial/Meritocratic/Japanese in the "unfixable, no ladder to lower" bucket. The shipped code is fine — the helper self-gates and correctly no-ops for them — so this is just worth correcting in the PR body so the scope note matches reality.

**Worth an in-game check:** a nomad at `nomadic_authority_5` losing an independence-faction war should drop to tier 4, and an Administrative ruler should now lose an `imperial_bureaucracy` tier.

common/scripted_effects/00_realm_effects.txt
```diff
@@ -165,7 +165,12 @@ decrease_nomadic_authority_effect = {
 	if = {
 		limit = { has_realm_law_flag = uses_nomadic_authority }
 		#Laws
+		#Unop Handle the top tier too, otherwise rulers at maximum nomadic authority lose nothing
 		if = {
+			limit = { has_realm_law = nomadic_authority_5 }
+			add_realm_law = nomadic_authority_4
```
pharaox, line 171: **reviewer:** Confirmed this resolves the Review 1 `Medium`, and the shape is right — prepending the tier-5 branch and demoting the old first branch to `else_if` keeps the chain mutually exclusive and still steps exactly one tier.

Two things I checked that make this safe: vanilla's own `09_mpo_wars.txt:1533-1538` switch opens with the identical `nomadic_authority_5 -> 4` transition (corroborating that `add_realm_law` replaces rather than stacks), and this helper has **zero vanilla callers**, so the added branch can't regress anything.

Also worth noting for future readers: `nomadic_authority` is the *only* one of the four ladders with five tiers (crown/tribal/imperial_bureaucracy are all 0-3), which is exactly why this was the sole helper with a top-tier gap. Stopping at tier 1 is correct — there is no `nomadic_authority_0`.

common/casus_belli_types/00_civil_war.txt
```diff
@@ -0,0 +1,2618 @@
...
+					any_character_struggle = {
+						phase_has_catalyst = catalyst_populist_uprise
+						involvement = involved
+					}
+				}
+				every_character_struggle = {
+					involvement = involved
+					activate_struggle_catalyst = {
+						catalyst = catalyst_populist_uprise
+						character = scope:defender #Unop `this` is the struggle here, not the character
```
pharaox, line 811: **reviewer:** Correct fix. Inside `every_character_struggle` the `this` scope is the **struggle**, but `activate_struggle_catalyst`'s `character` argument expects a character — so vanilla was passing the wrong type here.

Vanilla's own precedent is the same construct: `county_on_actions.txt:22-27` iterates `every_character_struggle` and passes `character = prev`. `scope:defender` is the equivalent target here and is more explicit than `prev`. Good that this was fixed rather than suppressed in `ck3-tiger.conf`.

PR #626, opened 2026-08-01
Grant treatment modifier on risky wound treatment critical success

### LINKED ISSUE #613: Risky wound treatment critical success downgrades the wound without granting a protective health modifier
**Reported on Steam by `Nider001` on `2026-07-27 16:22`** (<https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329>).

> The risky critical success outcome for wound treatments ("add_wound_treatment_modifier_effect" in "\scripted_effects\20_health_effects.txt") downgrades tier 2 or 3 wounds without granting a protective health modifier. This makes it almost strictly worse than a normal success, because characters can then die from the remaining health penalty, which seems like an oversight. Because of how the effect is structured, I believe the wound should instead be fully removed on critical success.

### LINKED ISSUE #614: Risky treatment regular success block in add_wound_treatment_modifier_effect is unreachable due to duplicate flag:critical_success limit
**Reported on Steam by `Nider001` on `2026-07-27 16:22`** (<https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329>).

> Also, "#Risky treatment critical success" is followed by "#Risky treatment regular success" in the same script, but the latter is redundant and can never be reached, because both "else_if" blocks have the same "limit". I believe the second "limit" should use "flag:success" instead of "flag:critical_success".

pharaox:
Fixes #613

**fixer:** The risky-treatment **critical success** branch of `add_wound_treatment_modifier_effect` only downgraded the wound tier and granted no modifier, while the regular-success branch granted `risky_wound_treatment_success_modifier`. A critical success was therefore *worse* than a regular one.

Added the same modifier at the end of the critical-success branch, guarded on a *residual* wound remaining — a `wounded_1` is fully removed by the downgrade, so there is no penalty left to negate and granting one would be a gratuitous buff.

Net health outcomes (`negate_health_penalty_add = 2` offsets a penalty rather than adding health, so it cannot over-heal; trait penalties `wounded_1 = -1`, `wounded_2 = -2`, `wounded_3 = -4`):

| Wound | Critical success | Regular success |
|---|---|---|
| `wounded_1` | removed → **0** | -1 + 2 → **0** |
| `wounded_2` | → `wounded_1` (-1) + 2 → **0** | -2 + 2 → **0** |
| `wounded_3` | → `wounded_2` (-2) + 2 → **0** | -4 + 2 → **-2** |

Critical is now always at least as good as regular, which was the invariant vanilla broke.

Chose the minimal remedy (keep the tier downgrade, add the modifier) over removing the wound outright: it preserves the "so you can see the trait you've lost" behaviour the vanilla comment calls out, and avoids turning a -4 straight to 0, which would read as rebalancing. The safe-treatment branch is correct and untouched.

**Note:** #614 (the `flag:success` half) is **already fixed on `main`** by 5ba07d43, so it is no longer part of this PR.

### DIFF
--- a/common/scripted_effects/20_health_effects.txt
+++ b/common/scripted_effects/20_health_effects.txt
@@ -3222,13 +3222,23 @@ add_wound_treatment_modifier_effect = {
 					remove_trait = wounded_3
 					add_trait = wounded_2
 				}
+				#Unop Also grant the treatment modifier, otherwise a critical success is worse than a regular one
+				if = {
+					limit = {
+						OR = {
+							has_trait = wounded_1
+							has_trait = wounded_2
+						}
+					}
+					add_character_modifier = { modifier = risky_wound_treatment_success_modifier days = wound_treatment_success_duration }
+				}
 			}
 
 			#Risky treatment regular success
 			else_if = {
 				limit = {
 					scope:treatment_result_treatment = flag:risky
-					scope:treatment_result_outcome = flag:critical_success
+					scope:treatment_result_outcome = flag:success #Unop Was critical_success, making this branch unreachable
 				}
 				add_character_modifier = { modifier = risky_wound_treatment_success_modifier days = wound_treatment_success_duration }
 			}
--- a/common/scripted_effects/20_health_effects.txt
+++ b/common/scripted_effects/20_health_effects.txt
@@ -3222,6 +3222,16 @@ add_wound_treatment_modifier_effect = {
 					remove_trait = wounded_3
 					add_trait = wounded_2
 				}
+				#Unop Also grant the treatment modifier, otherwise a critical success is worse than a regular one
+				if = {
+					limit = {
+						OR = {
+							has_trait = wounded_1
+							has_trait = wounded_2
+						}
+					}
+					add_character_modifier = { modifier = risky_wound_treatment_success_modifier days = wound_treatment_success_duration }
+				}
 			}
 
 			#Risky treatment regular success


common/scripted_effects/20_health_effects.txt
```diff
@@ -3222,13 +3222,23 @@ add_wound_treatment_modifier_effect = {
...
+					add_character_modifier = { modifier = risky_wound_treatment_success_modifier days = wound_treatment_success_duration }
+				}
 			}
 
 			#Risky treatment regular success
 			else_if = {
 				limit = {
 					scope:treatment_result_treatment = flag:risky
-					scope:treatment_result_outcome = flag:critical_success
+					scope:treatment_result_outcome = flag:success #Unop Was critical_success, making this branch unreachable
```
pharaox, line 3241: **reviewer:** `High` — this exact line is **already fixed on `main`** by `5ba07d43`, which reads:

```
					scope:treatment_result_outcome = flag:success #Unop: Was also checking critical_success
```

This branch predates that commit, so re-applying it here produces a merge conflict (confirmed via `git merge-tree`). Please rebase onto current `main` and drop this hunk — the PR should carry only the critical-success modifier addition above. Issue #614 is then already resolved and can be closed against `5ba07d43`.

pharaox (commented):
**reviewer:** Needs work — the #613 half is sound, but the #614 half is **already on `main`** and this branch conflicts with it.

**High — duplicate of an already-merged commit, and it will not merge.** `5ba07d43` *"Fix wrong flag check in add_wound_treatment_modifier_effect"* (ProZeratul, 2026-07-27) already set l.3231 to `flag:success`. This branch is based on a **stale `main`** (missing `5ba07d43` and `51ac8a9e`) and re-applies the same one-line change with different comment text. `git merge-tree` confirms a **real conflict** on exactly that line.

Suggested remedy:
1. Rebase onto current `main`.
2. **Drop the `flag:success` hunk** — keep only the critical-success modifier change (the #613 half).
3. Close **#614** against `5ba07d43` rather than this PR.

**The #613 half checks out and should survive the rebase unchanged:**

- Arithmetic verified against `00_health_modifiers.txt:88-91` (`negate_health_penalty_add = 2`) and trait penalties -1/-2/-4; the critical ≥ regular invariant is genuinely restored.
- **The residual-wound guard is well-judged.** I specifically checked the gangrene interaction — `gangrene_modifier` is already removed for risky treatments at l.3163-3173, *before* this branch — so the guard is not skipping a residual non-trait penalty.
- **Sibling system agrees:** `add_disease_treatment_modifier_effect` pairs `flag:critical_success` (l.2530) / `flag:success` (l.2608) on separate branches, each granting a modifier.
- Both call sites live: `health_events.txt:10031` and `:10108`.
- Choosing (b) over the reporter's (a) is right — preserves the l.3217 "so you can see the trait you've lost" behaviour and avoids taking `wounded_3` from -4 to 0.

**Tiger:** this branch reports `warning: 1`, but it is a `scope:wealth` warning in `01_ep1_court_artifact_creation_effects.txt`, unrelated to this change and an artefact of the stale base — a clean `main` baseline reports `No problems found`. It should vanish on rebase; worth re-checking then.

pharaox:
**fixer:** Re-fix addressing the High review finding.

- **Rebased onto current `main`** (was based on a stale `main` missing `5ba07d43` and `51ac8a9e`, which caused a genuine merge conflict).
- **Dropped the `flag:success` hunk.** Verified it is already on `main` at l.3231 (`#Unop: Was also checking critical_success`, commit `5ba07d43`). This PR now carries only the #613 half, and the `Fixes #614` line is removed — **#614 should be closed against `5ba07d43`, not this PR.**
- The Low finding is resolved as a side effect: no second `#Unop` marker is reintroduced on that line.

The #613 change itself is unchanged from Fix 1, since the review verified it as sound. Re-verified the underlying values directly against vanilla: `risky_wound_treatment_success_modifier` = `negate_health_penalty_add = 2` (`00_health_modifiers.txt:88-91`), and `wounded_1/2/3` = `-1/-2/-4` (`00_traits.txt:5562/5597/5636`), so the outcome table holds.

**Tiger:** `No problems found` — the `01_ep1_court_artifact_creation_effects.txt` warning was an artefact of the stale base and is gone after the rebase, as the review predicted.

pharaox (commented):
**reviewer:** Looks good — the rebase resolved both Review 1 findings, and the remaining change is minimal, correctly scoped, and tiger-clean.

Verified this cycle:

- **Stale base resolved.** `HEAD..origin/main` is empty and `git merge-tree` merges cleanly. The `flag:success` hunk is gone; l.3241 is now `main`'s merged version, not a re-application. No duplicate `#Unop` marker.
- **Guard placement correct.** The new `if` runs after the downgrade chain and inside the critical-success `else_if`, so it reads the *residual* trait and cannot leak into the safe or regular-success branches.
- **Preview/actual mismatch ruled out.** `add_wound_treatment_modifier_effect` is also called inside `show_as_tooltip` via `wound_treatment_result_tooltip_effect`, where `remove_trait`/`add_trait` don't mutate state — a post-downgrade `has_trait` guard would mispredict there. Both call sites of that wrapper (`20_health_effects.txt:3998`, `health_events.txt:8965`) pass `OUTCOME = failure`, which never reaches this branch, and `health.3105` calls the real effect in `immediate` with no preview. Not reachable.
- **Key assumption now corroborated.** `negate_health_penalty_add` is offset semantics, not surplus health: vanilla loc renders it as *"Negate Health Penalty"*; it is used as a permanent domicile-building modifier at values like 0.03-0.5; and the modifier's own description reads *"has reduced the severity of this character's wound."* The `wounded_1` exclusion is therefore right — no residual penalty to negate.
- **Tiger:** `No problems found`. The warning seen on the stale base is gone, confirming it wasn't attributable to this fix.

Style, indentation (tabs, matches vanilla), and commit hygiene all check out — one commit, one file, +10/-0.

**Merge note:** #614 is already fixed on `main` by `5ba07d43` and should be closed against that commit, not this PR.

common/scripted_effects/20_health_effects.txt
```diff
@@ -3222,6 +3222,16 @@ add_wound_treatment_modifier_effect = {
...
+				#Unop Also grant the treatment modifier, otherwise a critical success is worse than a regular one
+				if = {
+					limit = {
+						OR = {
+							has_trait = wounded_1
+							has_trait = wounded_2
+						}
+					}
+					add_character_modifier = { modifier = risky_wound_treatment_success_modifier days = wound_treatment_success_duration }
+				}
```
pharaox, line 3234: **reviewer:** Confirmed correct. This sits after the downgrade chain and inside the critical-success `else_if`, so `has_trait` reads the *residual* wound — `wounded_1` (fully removed) is correctly excluded as it has no penalty left to negate.

I specifically checked the tooltip hazard here: this effect is also invoked inside `show_as_tooltip` via `wound_treatment_result_tooltip_effect`, where `remove_trait`/`add_trait` don't actually mutate state, so a post-downgrade `has_trait` guard would read pre-downgrade traits and predict the wrong outcome. Both call sites of that wrapper pass `OUTCOME = failure`, so this branch is never reached through the preview path — no mismatch is possible.

PR #629, opened 2026-08-01
Fix Germania being uncreatable when the HRE already exists

### LINKED ISSUE #620: Empire of Germania cannot be created after the HRE is formed by a French ruler, despite both coexisting
**Reported on Steam by `Nider001` on `2026-07-29 12:48`** (<https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329>).

> If a king/emperor of France forms the HRE, the Empire of Germania will not be absorbed into it, allowing both to coexist on the de jure map. However, Germania's creation is still blocked in that case. Rulers are prompted to form the HRE instead, but it already exists. Since HRE is unlikely to lose all de jure land other than via restoring the true Roman Empire (already covered by previous fix), I suggest relaxing Germania's restrictions further by allowing it to be created after the decision to restore HRE gets used. Perhaps by mirroring the fix to Samhan/Goryeo empire creation restrictions.

pharaox:
Fixes #620

**fixer:** One line added to the existing `#Unop` `OR` in `e_germany.can_create`:

```
exists = title:e_hre.holder #Unop The HRE already exists, so it can no longer be formed instead
```

**Root cause** — the absorption is conditional but the block is not. `restore_holy_roman_empire_decision.is_valid` accepts `k_east_francia`, `k_bavaria`, `k_france`, `k_italy`, `k_lotharingia`, but the Germania de jure absorption in `restore_holy_roman_empire_decision_scripted_effect` only fires for `k_east_francia`/`k_bavaria`/`k_lotharingia`/`k_frisia`. A founder qualifying *only* via `k_france` or `k_italy` forms the HRE without absorbing Germania, so both empires coexist. Germania then stays blocked for Catholics ("form the HRE instead"), while the HRE can never be re-formed — `is_shown` contains `NOR = { exists = title:e_hre.holder }`. Permanent dead end. The existing #440 escape hatch doesn't help: `flag_restored_roman_empire` is only set by the Roman path.

**Why this gate, released this way.** The trigger's own loc states the intent outright — *"As a [Faith] ruler you can reform the $e_hre$ instead"*. That sentence is simply false once the HRE exists, so releasing the gate on exactly that condition targets the real intent rather than widening the proxy. It mirrors the #440 fix already sitting in this same `OR`.

**Rejected alternatives:**
- *Goryeo-style "decision consumed" test* — `restore_holy_roman_empire_decision` does **not** register in `unavailable_unique_decisions`; it only sets a global guarded by `is_ai = no`, so copying that pattern would silently fail when the AI forms the HRE.
- *Widening the absorption `limit`* — would prevent the coexistence entirely, but is a much larger change to the de jure map, and the reporter prefers coexistence to remain.

**Trade-off:** this also unblocks Germania for a Catholic when an HRE *did* absorb it — but in that case Germania has no de jure kingdoms left, so creation is inert and still governed by the normal de jure/tier requirements.

Note: the analysis flagged `form_germania_christian_trigger` as having no localization. It resolves via `common/trigger_localization/00_landed_title_triggers.txt` to `FORM_GERMANIA_CHRISTIAN_TRIGGER`, which does exist — no loc work needed.

### DIFF
--- a/common/landed_titles/00_landed_titles.txt
+++ b/common/landed_titles/00_landed_titles.txt
@@ -4566,6 +4566,7 @@ e_germany = {
 					}
 				}
 				exists = global_var:flag_restored_roman_empire
+				exists = title:e_hre.holder #Unop The HRE already exists, so it can no longer be formed instead
 			}
 		}
 		rule_title_creation_imperial_power_projection_title_creation_trigger = yes


common/landed_titles/00_landed_titles.txt
```diff
@@ -4566,6 +4566,7 @@ e_germany = {
 					}
 				}
 				exists = global_var:flag_restored_roman_empire
+				exists = title:e_hre.holder #Unop The HRE already exists, so it can no longer be formed instead
```
pharaox, line 4569: **reviewer:** Confirmed correct, and the `.holder` suffix is load-bearing — worth recording why.

`e_hre` has no `history/titles/e_hre.txt`, so the title is **defined but unheld** at game start. `exists = title:e_hre` would therefore evaluate true from day one and unblock Germania for every Catholic immediately, defeating the gate entirely. `exists = title:e_hre.holder` is the correct "actually exists" test and matches the dominant vanilla idiom, including the HRE decision'''s own `is_shown` (`80_major_decisions_middle_europe.txt:704`) that this condition is the complement of.

Note the `k_east_francia.can_create` precedent cited in the fix write-up uses the bare `NOT = { exists = title:e_hre }` (l.6949) — for an AI-only branch where the distinction is less consequential. Good that it was not copied here.

**In-game check worth doing:** that the gate still bites for a Catholic *before* any HRE exists, and that Germania becomes creatable once a French-founded HRE coexists with an unabsorbed `e_germany`.

pharaox (commented):
**reviewer:** Looks good — minimal, correctly targeted, and it picks the right form of the existence test.

Verified beyond the diff:

- **The gate now matches its own stated intent.** `FORM_GERMANIA_CHRISTIAN_TRIGGER` reads *"As a [faith] ruler you can reform the $e_hre$ instead"* — false once the HRE exists. Releasing on exactly that condition targets the real condition instead of widening a proxy.
- **`.holder` is the right choice, and notably better than the precedent quoted in the write-up.** There is no `history/titles/e_hre.txt`, so `e_hre` is defined-but-unheld at game start — `exists = title:e_hre` would be **true from day one** and unblock Germania for every Catholic immediately. `exists = title:e_hre.holder` is also the dominant vanilla idiom (the HRE decision itself at l.704, `80_major_decisions_roman.txt:1173`, `game_start.txt`, steward tasks, artifact events). The `k_east_francia.can_create` precedent cited actually uses the *bare* form — good that it was not copied.
- **Exact logical complement of the blocking gate:** `restore_holy_roman_empire_decision.is_shown` blocks on `NOR = { exists = title:e_hre.holder }`. No gap, no overlap.
- **Checked the #440 trap** — no other guard in the same `can_create` still blocks the case; the only sibling is the game-rule power-projection trigger, which is unrelated.
- **Root cause asymmetry re-verified:** `is_valid` accepts `k_france`/`k_italy` (l.741-752) but the de jure absorption is gated on `k_east_francia`/`k_bavaria`/`k_lotharingia`/`k_frisia` only (`00_major_decisions_scripted_effects.txt:3155-3172`).
- **No regression when absorption did happen** — Germania is stripped of de jure kingdoms, so creation stays inert under the normal de jure majority requirement.
- **Correctly ruled out the Goryeo pattern** (the decision does not register in `unavailable_unique_decisions`, and its only global write is `is_ai = no`-guarded), and the retraction of the "missing loc" claim is correct — I confirmed the key resolves via `00_landed_title_triggers.txt:153`.

No findings. One in-game check noted inline.

PR #641, opened 2026-09-20
Fix is_of_major_interest_trigger

pharaox:
Fix `is_of_major_interest_trigger` to always properly check that the scoped character is of major interest to `$CHARACTER$`, instead of vice versa. 

The vanilla version has flipped two of the checks, resulting in weird effects when this trigger is used. Notably, it checks if the scoped character has `$CHARACTER$` as liege, instead of vice versa.


common/scripted_triggers/00_relation_triggers.txt
```diff
@@ -0,0 +1,1230 @@
...
+		}
+	}
+}
+
+is_of_major_interest_trigger = {
+	NOT = { this = $CHARACTER$ }
+	OR = {
+		is_heir_of = $CHARACTER$
+		is_diarch_of_target = $CHARACTER$
+		$CHARACTER$.liege ?= this #Unop Check if I am the character's liege, not they are my liege
```
pharaox: Confirmed correct, and worth flagging for testers as the visible behavioral delta.

Before: a ruler's **liege** counted as major interest — but that case was already covered by `is_of_minor_interest_trigger`'s liege-or-above branch, so it was largely redundant.
After: a ruler's **vassals** count as major interest, which is what the minor trigger's `NOT = { this = $CHARACTER$.liege }` exclusion implies was always intended.

This shifts character-importance weighting, court-trait message selection and prison memory buffering. Intended, but the thing to watch in-game.

common/scripted_triggers/00_relation_triggers.txt
```diff
@@ -0,0 +1,1230 @@
...
+			this = $CHARACTER$.house.house_head
+		}
+		# Activity Intents
+		trigger_if = {
+			#Unop Check if I am the intent target of the character, not vice versa
+			limit = {
+				exists = $CHARACTER$.involved_activity
+				exists = $CHARACTER$.intent_target
+			}
+			$CHARACTER$.intent_target = this
```
pharaox: Correct per `event_targets.log`: *`intent_target` - Get the intent target of scoped character*. Vanilla asked "is `$CHARACTER$` the target of **my** intent", which is the same inversion as the liege branch.

Good that both `exists` guards were moved to `$CHARACTER$` alongside the body — the block stays internally consistent, mirroring vanilla's all-`this` shape rather than becoming a mixed-scope check.

pharaox (commented):
**reviewer:** Looks good — the flip direction is correct and the pattern class is fully enumerated.

Verified the trigger's contract from its call sites rather than from the diff alone:

- `01_character_weighting_values.txt` calls `is_of_major_interest_to_root_trigger` beside `is_heir_of = root` / `is_consort_of = root` / `is_child_of = root` — all "`this` is X **of** root". `00_ep1_court_type_effects.txt` filters `random_courtier` (already root's courtier) by it, to rank *root's own* courtiers. Contract: **`this` is of major interest to `$CHARACTER$`**. Both patched branches now match it.
- The liege fix is independently corroborated by the sibling `is_of_minor_interest_trigger`, which covers liege-or-above but *explicitly excludes* `this = $CHARACTER$.liege` — an exclusion that only makes sense if the direct-liege case belongs to the **major** trigger. Vanilla duplicated the minor case and left the major one uncovered; this closes exactly that gap.
- Checked every other branch for the same flip: `is_heir_of`, `is_child_of`, `is_diarch_of_target`, `is_councillor_of`, `is_spouse_of` are all "scoped character is X of target" per `triggers.log`, and `religious_head` / `house_head` already use the `$CHARACTER$.`-prefixed form this fix applies. No sibling flip left behind.
- `make tiger` re-run on the branch: clean (0/0/0/0/0). No conf suppressions. Vanilla file added in its own commit, unmodified; diff vs `${CK3_BASE_DIR}` shows only the two intended changes.

No findings. One note for in-game testing below.

pharaox:
<details><summary>🤖 handover document</summary>

# PR 641 — Fix is_of_major_interest_trigger

Intent (from PR description, no linked issue): `is_of_major_interest_trigger` should check that the
scoped character is of major interest **to `$CHARACTER$`**. Vanilla has two branches flipped —
notably it checks whether the scoped character has `$CHARACTER$` as liege, instead of vice versa.

## Review 1

**Verdict:** Looks good

The trigger's contract is unambiguous from its call sites, and the fix matches it.

**Evidence gathered:**

- **Contract confirmed by callers.** `common/script_values/01_character_weighting_values.txt`
  calls `is_of_major_interest_to_root_trigger` alongside `is_heir_of = root`, `is_consort_of = root`,
  `is_child_of = root` — all of the form "`this` is X *of* root". Same in
  `00_ep1_court_type_effects.txt`, where `random_courtier` (already root's courtier) is filtered by
  `is_of_major_interest_to_root_trigger` to rank *root's own courtiers* by importance to root.
  So the contract is: **`this` is of major interest to `$CHARACTER$`**.

- **Liege branch (line 573).** Vanilla `liege ?= $CHARACTER$` asks "is `$CHARACTER$` my liege" —
  backwards. Corroborated by the sibling `is_of_minor_interest_trigger`, which covers
  liege-or-above via `$CHARACTER$ = { target_is_liege_or_above = prev }` and *explicitly excludes*
  `this = $CHARACTER$.liege`. That exclusion only makes sense if the direct-liege-of-`$CHARACTER$`
  case is meant to live in the **major** trigger. Vanilla therefore both duplicated the minor case
  and left the major case (my vassal / `$CHARACTER$`'s liege is me) uncovered. The fix
  (`$CHARACTER$.liege ?= this`) closes exactly that gap.

- **Activity intent branch (lines 609-616).** `event_targets.log`: `intent_target - Get the intent
  target of scoped character`. Vanilla `intent_target = $CHARACTER$` asks "is `$CHARACTER$` the
  target of *my* intent" — backwards for the same reason. The fix flips both `exists` guards and the
  body to `$CHARACTER$`, keeping the block internally consistent (guards and body reference the same
  scope, exactly as vanilla did, just on the other side).

- **Pattern class enumerated.** Checked every other branch of the trigger against the same contract:
  `is_heir_of`, `is_child_of`, `is_diarch_of_target`, `is_councillor_of`, `is_spouse_of` are all
  "scoped character is X of target" per `triggers.log` — correct as-is. The `religious_head` and
  `house_head` branches already use the `$CHARACTER$.`-prefixed form the fix applies here, so the
  two patched branches were the only outliers. No sibling flip left unfixed.

- **Tiger clean.** `make tiger` re-run on the branch: `fatal: 0, error: 0, warning: 0, untidy: 0,
  tips: 0`. No `ck3-tiger.conf` suppressions added.

- **Commit structure.** Two commits: `02825ba4 Add base game files` (vanilla file unmodified, 1229
  lines) then `8e22ef07 Fix is_of_major_interest_trigger` (5 insertions, 4 deletions). Matches the
  project convention. Diff against `${CK3_BASE_DIR}` confirms the only deviations from vanilla are
  the two intended changes.

- **Minimality.** 2 branches touched, no new state, no new triggers, surrounding formatting matched,
  `#Unop` markers present and accurate.

**Findings:** none.

**Note for testers:** this trigger feeds character-importance weighting, court-trait message
selection, prison memory buffering and tournament qualification. The liege branch changes which
characters qualify — after the fix a ruler's **vassals** count as major interest (previously their
**liege** did, which the minor trigger already covered). That is the intended correction, but it is
the visible behavioral delta worth watching in-game.
</details>

PR #642, opened 2026-09-20
Fix house_has_feud_relation_with_trigger

pharaox:
Fix `house_has_feud_relation_with_trigger` to ensure it can be invoked with relative links (`liege`, `prev`) etc. There are 3 call sites in the base game using it with such links, potentially more in mods. They produce errors similar to:

```
Error: Event target link 'liege' did not get a matching scope type. Expected 'character', but got 'dynasty_house'
```


common/scripted_triggers/03_bp1_scripted_triggers.txt
```diff
@@ -13,11 +13,12 @@ house_has_feud_relation_with_trigger = {
 	exists = house
 	exists = $TARGET$.house
 	house != $TARGET$.house
+	$TARGET$.house = { save_temporary_scope_as = temp_target_house } #Unop Use saved scope to avoid issues with relative links
```
pharaox: Correct and well-placed — `exists = $TARGET$.house` on L14 guarantees this can never run on a missing house.

Worth noting for the record: vanilla does exactly this two hundred lines below, in `house_feud_ai_murder_start_trigger` (L224-234), which saves `instigator_temp` up front and then uses `scope:instigator_temp.house` inside the same `any_house_relation` → `any_relation_house` nesting. Good idiom match, and `temp_target_house` collides with nothing in vanilla or the mod.

**Low:** the trigger now leaks `scope:temp_target_house` to callers as a side effect. Unavoidable with this idiom and vanilla does the same, so not worth changing — just keep the distinctive name if this ever gets refactored.

common/scripted_triggers/03_bp1_scripted_triggers.txt
```diff
@@ -13,11 +13,12 @@ house_has_feud_relation_with_trigger = {
...
 	OR = {
 		house = {
 			any_house_relation = {
 				has_house_relation_level = feud
-				any_relation_house = { $TARGET$.house = this }
+				any_relation_house = { scope:temp_target_house = this } #Unop Use saved scope to avoid issues with relative links
```
pharaox: This is the line that was actually broken: `$TARGET$` was re-resolved here, where the scope is a `dynasty_house`.

**Low / follow-up only:** with the fix in the trigger, the three existing call-site workarounds become redundant — `ep3_emperor_yearly_2.txt:4978-4979` (`unop_liege`), `ep3_powerful_families_8.txt:1367,1401` (`scope:house_head_temp`), and `10_tgp_test_interactions.txt:72`. They still behave correctly, since they pass a `scope:` which this handles identically. Leaving them out of this PR is the right call for a minimal change; flagging only as possible cleanup later.

pharaox (commented):
**reviewer:** Looks good — right diagnosis, right altitude, and the idiom has vanilla precedent 210 lines below the fix.

**Mechanism verified.** Macro params are textual substitution, so vanilla's `any_relation_house = { $TARGET$.house = this }` re-resolves `$TARGET$` at a depth where the scope is a `dynasty_house`. Relative links have no meaning there — hence the reported error. Resolving once at the top, where the caller's scope is still intact, is the correct remedy.

**Reproduced at a real call site.** `ep3_powerful_families_8.txt:1366` calls `TARGET = prev` from inside `any_powerful_family = { house_head = { ... root = { ... } } }`; `prev` is the `house_head` character, re-evaluated against a `dynasty_house`. That is the error verbatim.

**Call sites enumerated.** Exactly the 3 relative-link sites the description claims (`liege` ×1 at `ep3_emperor_yearly_2.txt:4975`, `prev` ×2 at `ep3_powerful_families_8.txt:1366,1399`). The other 100+ pass `root`/`scope:`/`title:`, which are absolute and resolve identically at any depth — so the change is behavior-preserving for them. The only delta is resolving `$TARGET$.house` once rather than per-iteration, same result for absolute targets.

**Approach corroborated three ways:**
1. Vanilla `house_feud_ai_murder_start_trigger` (same file, L224-234) solves the identical problem identically — `save_temporary_scope_as` up front, then `scope:instigator_temp.house` inside the very same `any_house_relation` → `any_relation_house` nesting.
2. The `<derived-scope> = { save_temporary_scope_as = X }` form is attested across many vanilla scripted triggers (`00_religious_triggers`, `00_secret_triggers`, `06_fp3_scripted_triggers`, …).
3. Unop already carries **three hand-rolled workarounds for this exact bug** (`ep3_emperor_yearly_2.txt:4978` saving `unop_liege`; `ep3_powerful_families_8.txt:1367,1401`; `10_tgp_test_interactions.txt:72`). Moving the fix into the trigger is the right altitude — one change covers all call sites, including modded ones.

**Safety checks:** `exists = $TARGET$.house` precedes the save, so it can't run on a missing house; `house != $TARGET$.house` still evaluates on the caller's scope; no collision — `temp_target_house` appears nowhere else in vanilla or the mod. `make tiger` re-run: clean (0/0/0/0/0), no conf suppressions.

Two `Low` notes inline, neither blocking. Suggest in-game validation via observer mode: the `did not get a matching scope type` errors should disappear from `error.log`.

pharaox:
<details><summary>🤖 handover document</summary>

# PR 642 — Fix house_has_feud_relation_with_trigger

Intent (from PR description, no linked issue): make `house_has_feud_relation_with_trigger` safe to
invoke with relative links (`liege`, `prev`). Vanilla re-evaluates `$TARGET$.house` deep inside
nested iterators where the scope is a `dynasty_house`, producing:
`Error: Event target link 'liege' did not get a matching scope type. Expected 'character', but got 'dynasty_house'`.

## Review 1

**Verdict:** Looks good

**Mechanism confirmed.** Vanilla's inner check is
`any_relation_house = { $TARGET$.house = this }`. Macro parameters are textual substitution, so
`$TARGET$` is re-resolved at that depth — inside `house = { any_house_relation = {
any_relation_house = { ... } } }`, where the scope is a `dynasty_house`. A relative link like
`liege` or `prev` has no meaning there, hence the reported error. The fix resolves `$TARGET$.house`
once at the top of the trigger (where the caller's scope is still intact) and compares against the
saved scope inside the iterator. Correct diagnosis, correct remedy.

**Bug reproduced at a real vanilla call site.** `ep3_powerful_families_8.txt:1366` calls it as
`TARGET = prev` from inside `any_powerful_family = { house_head = { ... root = { ... } } }`. There
`prev` is the `house_head` *character*; re-evaluated inside `any_relation_house` it resolves against
a `dynasty_house`. This is the error verbatim.

**Call sites enumerated.** All vanilla call sites using a relative link — exactly the 3 the PR
claims:
- `events/dlc/ep3/ep3_emperor_yearly_2.txt:4975` — `TARGET = liege`
- `events/dlc/ep3/ep3_powerful_families_8.txt:1366` — `TARGET = prev`
- `events/dlc/ep3/ep3_powerful_families_8.txt:1399` — `TARGET = prev`

Every other call site (100+) passes `root`, a `scope:` or a `title:`, which are absolute and
re-resolve identically at any depth — so the fix is behavior-preserving for them. The only
semantic change is that `$TARGET$.house` is now resolved once instead of per-iteration, which for
absolute targets yields the same house.

**Approach is the vanilla-sanctioned idiom, with precedent in the same file.**
`house_feud_ai_murder_start_trigger` (vanilla, `03_bp1_scripted_triggers.txt:224-234`, ~210 lines
below the fix) solves the identical problem the identical way: `save_temporary_scope_as =
instigator_temp` up front, then `scope:instigator_temp.house` inside the very same
`any_house_relation` → `any_relation_house` nesting. `has_RelationToMe_relation` and
`get_characters_of_major_interest_trigger` do the same. The
`<derived-scope> = { save_temporary_scope_as = X }` form is attested across many vanilla scripted
triggers (`00_religious_triggers.txt`, `00_secret_triggers.txt`, `06_fp3_scripted_triggers.txt`,
etc.). The `_temp`-style name matches local convention.

**Diagnosis independently corroborated by three pre-existing Unop workarounds** for this exact bug,
each of which hand-rolls the fix at the call site:
- `events/dlc/ep3/ep3_emperor_yearly_2.txt:4978-4979` — `liege = { save_temporary_scope_as =
  unop_liege }` then `TARGET = scope:unop_liege`
- `events/dlc/ep3/ep3_powerful_families_8.txt:1367,1401` — `TARGET = scope:house_head_temp`
  *"Use saved scope since prev changes in the trigger"*
- `common/character_interactions/10_tgp_test_interactions.txt:72` — *"Use correct scope"*

This PR moves the fix into the trigger, which is the right altitude: one change covers all call
sites including modded ones, instead of N call-site patches. The existing workarounds stay correct
(they pass a `scope:`, which the fix handles identically) — merely redundant. Leaving them is
consistent with "prefer the minimal change".

**Ordering / safety checks:**
- `exists = $TARGET$.house` is asserted on line 14, before the save on line 16, so the save can
  never run on a missing house.
- `house != $TARGET$.house` (line 15) still runs on the caller's scope, unaffected.
- `save_temporary_scope_as` in a trigger context does not short-circuit evaluation (vanilla relies
  on this in the precedents above).
- No name collision: `temp_target_house` appears nowhere else in vanilla or the mod.

**Tiger clean.** `make tiger` re-run on the branch: `fatal: 0, error: 0, warning: 0, untidy: 0,
tips: 0`. No `ck3-tiger.conf` suppressions added.

**Minimality / style.** 2 lines, one file, no new state beyond the temp scope, `#Unop` markers
present and accurate on both changed lines, surrounding formatting matched. No vanilla file added
(the file was already in the mod), so the separate-commit rule does not apply.

**Findings:**

| Severity | Description |
|---|---|
| Low | The trigger now leaks `scope:temp_target_house` to callers as a side effect (unavoidable with this idiom, and vanilla does the same in `house_feud_ai_murder_start_trigger`). No collision exists today; worth keeping the distinctive name if the trigger is ever refactored. |
| Low | The three pre-existing Unop call-site workarounds are now redundant. Correct to leave them in this PR (minimal change), but they could be simplified in a follow-up. |

No `High` or `Medium` findings.

**Note for testers:** the fix is exercised by the Byzantine governor content
(`ep3_emperor_yearly_2`) and powerful-families estate content (`ep3_powerful_families_8`). Running
observer mode and confirming the `did not get a matching scope type` errors disappear from
`error.log` would validate it directly.
</details>

PR #657, opened 2026-09-21
Fix multi-slot court position dispatch never matching its stored value

### LINKED ISSUE #656: The event "...: the Bodyguard" doesn't appoint the character as Bodyguard after recruiting
The event "...: the Bodyguard" (`ep3_laamp_decision_event.1057`) lets you recruit a character with appoint them as bodyguard, but the latter may not happen in some cases;

* In `1057_hire_bodyguard_effect`, `court_position_grant_effect` is just a tooltip, with a comment `#The actual effect happens in after`.
* The code in after looks correct, but for whatever reason the character is not appointed as bodyguard, even though they meet the conditions (and can be appointed manually after).

Unop already fixes an issue in `can_appoint_char_to_court_position` that could cause this, and it fixed similar issues elsewhere, but for whatever reason doesn't seem to fix this one.

pharaox:
Fixes #656

**fixer:** The multi-slot dispatch in `court_position_grant_effect` stores the position *with* the `_court_position` suffix but compares *without* it:

```
save_scope_value_as = { name = unop_multi_slot_cp  value = flag:$POS$_court_position }  # -> flag:bodyguard_court_position
...
scope:unop_multi_slot_cp = flag:bodyguard    # never matches
```

So once the outer `unop_is_multi_slot_cp` gate correctly admits a multi-slot position, no inner arm can match. `unop_get_worst_cp_holder` is never called, `scope:unop_worst_char_for_cp` never exists, and `scope:old_holder` is never set — while the `else` that would otherwise set it is skipped, because the position legitimately took the multi-slot path.

With `old_holder` unset, the grant falls to the `else` branch and calls plain `appoint_court_position` instead of `replace_court_position`. When both bodyguard slots are already filled there is nowhere to put the recruit, so it silently no-ops — which is exactly the reported repro. Manual appointment works because the UI path replaces properly.

Fix: add the `_court_position` suffix to all four comparisons, matching the convention `unop_is_multi_slot_cp` already uses (suffix on both sides). Applied at **both** affected sites — `court_position_grant_effect` and `unop_get_worst_or_none_cp_holder_effect`, which carries the identical mismatch. Both came in with 268a2f70.

The `SLOT` arguments are correct and unchanged: bodyguard/lady_in_waiting/garuda are `max_available_positions = 2` and court_artificer is `3`, matching the 2-slot and 3-slot arms.

Scope of impact: only the four positions `unop_is_multi_slot_cp` covers, and only when full. The five sibling LAAMP recruit events are all single-slot and take the `else` branch, so they were never affected.

Please verify in-game with both bodyguard slots filled that the recruited character now replaces the worst existing bodyguard, and check `error.log` in observer mode.


pharaox:
**fixer:** Replaced the fix — you and `@analyzer` were right, my first attempt was based on the retracted `scope:liege` theory and would not have fixed anything. I'd pushed it before reading the correction; apologies for the noise.

Force-pushed the branch so it now carries a single commit with the real fix: adding the `_court_position` suffix to the four multi-slot comparisons, so the stored `flag:$POS$_court_position` value can actually match. Applied at both sites — `court_position_grant_effect` and `unop_get_worst_or_none_cp_holder_effect`.

I chose to add the suffix to the comparisons rather than drop it from the stored value, since that matches `unop_is_multi_slot_cp`'s own convention of carrying the suffix on both sides.

Two things I checked while verifying the chain:
- The `SLOT` arguments are already correct — bodyguard/lady_in_waiting/garuda are `max_available_positions = 2`, court_artificer is `3`, matching the 2-slot and 3-slot arms. No change needed there.
- No other `scope:unop_multi_slot_cp` comparison exists repo-wide, so these four (×2 sites) are the complete set.

`make tiger` is clean. The repro to confirm is the one from the issue: both bodyguard slots filled, so the grant must take the `replace_court_position` path.

pharaox:
**fixer:** Removed the `#Unop` comments — noted that they mark fixes to *vanilla* code, and these lines are Unop's own, so there was nothing to mark. Force-pushed; the diff is now purely the suffix addition. The only remaining `#Unop` is the pre-existing `minister_bureaucrat` reminder, which I left untouched.

`make tiger` still clean.

pharaox (commented):
**reviewer:** Looks good — no findings. This is the real root cause, and the rewrite cleanly abandons the retracted `scope:liege` theory.

I re-read the full issue thread (your `valid_character` docs refutation and the analyzer's retraction) and confirmed the branch now carries a **single** commit containing only the suffix change — no trace of the `save_temporary_scope_as = liege` edit remains.

**Failure chain traced end to end in-file:**

- `court_position_grant_effect:577-580` stores `value = flag:$POS$_court_position` → `flag:bodyguard_court_position`, while the arms compared against bare `flag:bodyguard`.
- No arm matches → `unop_get_worst_cp_holder` never runs → `scope:unop_worst_char_for_cp` never exists → the `if` at :604-607 never sets `old_holder`, and the `else` at :609 that would otherwise set it is skipped because the position legitimately took the multi-slot path → the grant falls to the `else` at :643 and calls `appoint_court_position` instead of `replace_court_position`, a silent no-op when the position is full. Exactly the reported repro.

**The suffix direction is verifiably the house convention, not a coin flip:** `unop_is_multi_slot_cp` (`00_unop_triggers.txt:896-902`) — the very gate that admits these positions — already carries the suffix on *both* sides (`$CP$_court_position = flag:bodyguard_court_position`). This makes the inner branches consistent with the gate that lets them in.

Also verified:

- **The chain works once matched:** `unop_get_worst_cp_holder` (`00_unop_effects.txt:17-38`) fires only when `count >= $SLOT$` (position full) and saves the lowest-aptitude holder via `order_by` × -1 — so `old_holder` is set and `replace_court_position` is selected.
- **`SLOT` args against vanilla `max_available_positions`:** bodyguard 2 (`00_court_positions.txt:10765`), lady_in_waiting 2 (:7548), garuda 2 (:11721), court_artificer 3 (:11405). All four match their arms.
- **Completeness:** `scope:unop_multi_slot_cp` repo-wide returns exactly the 10 lines across your two sites, all now suffixed — no third site, no surviving bare comparison.
- **Scope-of-impact confirmed.** Every camp officer in `00_camp_officers.txt` is `max_available_positions = 1`, and `court_physician_court_position` is 1 — so the five sibling LAAMP events take the `else` at :609-611 and were never affected. Correct to retract the earlier "all six share the defect" conclusion and to reclassify origin as `mod`.
- **Omitting `#Unop` markers is right here:** `git log` confirms both blocks came in with `268a2f70`, so these are Unop's own lines, not vanilla — a vanilla-fix marker would mislead. Updating the commented `minister_bureaucrat` reminder in step with its siblings keeps it consistent for whenever it's enabled.
- **Tiger re-run independently:** `fatal: 0, error: 0, warning: 0, untidy: 0, tips: 0`. `git diff --check` clean.

The one thing static analysis can't settle is the key assumption that `replace_court_position` succeeds where `appoint_court_position` no-ops — worth the in-game check with **both bodyguard slots filled**, confirming the recruit replaces the lowest-aptitude bodyguard, plus an `error.log` pass.

PR #658, opened 2026-09-21
Fix Wanua governments being treated as Tribal in government dispatch

### LINKED ISSUE #655: After conquering a Wanua ruler as an adventurer you become Tribal instead of Wanua
After conquering a ruler as an adventurer you should adopt their government. For Wanua rulers however, you become Tribal, even though everyone around you is either Wanua or Mandala.

This is because `ep3_become_landed_transfer_effect` selects on `government_has_flag` and checks `government_is_tribal` before `government_is_wanua`. However, the Wanua government has both flags, so Tribal is with priority. Simply changing the order should fix the issue.

The effect should be checked for other similar cases.

pharaox:
Fixes #655

**fixer:** `wanua_government` declares **both** `government_is_tribal` and `government_is_wanua`. Three dispatch sites test the tribal flag first, so Wanua always matches the tribal arm and the wanua arm below is unreachable dead code — conquering a Wanua ruler as an adventurer turns you Tribal.

Vanilla already provides the disambiguator: `tribal_government` carries a dedicated `government_is_tribal_excluding_wanua` flag, which Wanua does not. This fix uses that flag in the tribal arm at all three sites rather than reordering the arms, because it makes the exclusion explicit instead of relying on implicit ordering that a future vanilla merge could silently reintroduce. It also mirrors the `NOT = { government_has_flag = government_is_wanua }` idiom vanilla already uses in `00_government_custom_loc.txt`.

Sites fixed:
- `07_dlc_ep3_scripted_effects.txt` — the `switch` on `scope:government_giver` (the reported path).
- `07_dlc_ep3_scripted_effects.txt` — a separate `if/else_if` chain in the same effect, where a Wanua **liege** hit the tribal branch. This is an independent path, so fixing the switch alone would not have covered it.
- `00_faction_effects.txt` — the identical `switch` in peasant-uprising government setup, whose downstream `scope:old_government ?= flag:wanua` test was likewise unreachable.

Exactly two governments carry `government_is_tribal` (tribal and wanua) and only tribal carries the narrower flag, so the substitution changes behaviour for Wanua only — no other government is affected.

Left alone deliberately: `ep3_admin_events.txt` (no wanua arm — a 3-way flavour split where tribal is the sensible fallback), `10_ach_effects.txt` (neither arm), and `00_government_custom_loc.txt` / `80_major_decisions.txt`, which already handle the overlap correctly.

Please verify in-game that conquering a Wanua ruler as an adventurer now yields Wanua, and check `error.log` in observer mode.


common/scripted_effects/07_dlc_ep3_scripted_effects.txt
```diff
@@ -873,7 +873,7 @@ ep3_become_landed_transfer_effect = {
 			government_is_clan = {
 				save_scope_value_as = { name = old_government value = flag:clan }
 			}
-			government_is_tribal = {
+			government_is_tribal_excluding_wanua = { #Unop: Wanua has the government_is_tribal flag too, so it matched this case first and the wanua case below was unreachable
```
pharaox: **reviewer:** `Low` (cosmetic, non-blocking) — this marker and its two siblings run ~110 chars and are near-identical. The project's style leans terser, and `#Unop` without the colon is the preferred form:

```suggestion
			government_is_tribal_excluding_wanua = { #Unop Wanua also has government_is_tribal, which made the wanua case below unreachable
```

Entirely optional — the content is accurate and the reason is the right one to record.

pharaox (commented):
**reviewer:** Looks good — the dedicated-flag remedy is correct, minimal, and verified effective at all three dispatch sites.

What I checked beyond the diff:

- **Flag declarations re-verified in vanilla.** Only `tribal_government` (:247) and `wanua_government` (:311) declare `government_is_tribal`; only `tribal_government` declares `government_is_tribal_excluding_wanua` (:248). So the substitution changes behaviour for Wanua alone — the blast-radius claim holds.
- **Unop's own `00_government_types.txt` override.** This was the load-bearing dependency, since Unop overrides that file and the fix would silently break if its copy lacked the flag. It carries `government_is_tribal_excluding_wanua` at the identical line; its only divergence from vanilla is unrelated (`ai = { use_legends = yes }`).
- **All three sites are genuinely effective**, not just syntactically changed:
  - Site 1 — the wanua arm at :901 is now reachable and `flag:wanua = { change_government = wanua_government }` (:1113) exists downstream.
  - Site 2 — a Wanua liege now falls through to the `## Wanua.` branch (:1288). The untouched `scope:old_government ?= flag:tribal` disjunct is correct by construction now that site 1 no longer stores `flag:tribal` for Wanua.
  - Site 3 — the `flag:wanua` disjunct at `00_faction_effects.txt:749` is now reachable, so a Wanua predecessor **outside** `world_asia_southeast_islands` correctly yields Wanua instead of falling through to Tribal. Previously only the geographic fallback saved it.
- **Pattern class independently re-enumerated.** Grepped every `trigger = government_has_flag` switch in vanilla `common/` and `events/`: exactly four, matching the analysis. The fourth (`ep3_admin_events.txt:53`) is correct by design — no wanua arm, so no unreachable arm, plus an explicit `feudal_gov` default. Not a missed sibling.
- **Regression vector.** `scope:old_government` is consumed only within the two changed files, so nothing else relied on Wanua being folded into `flag:tribal`.
- **Tiger re-run independently:** `fatal: 0, error: 0, warning: 0, untidy: 0, tips: 0`. No `ck3-tiger.conf` suppressions added. Single commit, two files, no `.claude/` leakage.

Agreed with choosing the dedicated flag over reordering — it makes the exclusion explicit at the point of test instead of leaving correctness dependent on implicit arm order, which a future vanilla merge could silently reintroduce.

Only a `Low` note on comment length (inline). Worth an in-game confirmation from the reporter that conquering a Wanua ruler as an adventurer now yields Wanua.

PR #659, opened 2026-09-21
Fix A Volunteer leaving a recruited guest at the original court

### LINKED ISSUE #653: The event "A Volunteer" may leave the volunteer visiting the original court if they are a guest
The event "A Volunteer" offers courtiers or guests of a landed ruler as "volunteers" to join a LAAMP's court. If a guest is chosen, they are added to the LAAMP's court but remain in the original court as a guest (a courtier guest rather than a pool guest), and therefore don't travel with the LAAMP.

See this video: https://www.youtube.com/watch?v=G-IIRQnhF3w, at 8:40 to 8:45.

pharaox:
Fixes #653

**fixer:** `laamp_base_contract_schemes.0321` picks its candidates with `ordered_courtier_or_guest`, so the recruit can be a *guest* of the landed ruler. The transfer effect does only `add_courtier`, which re-parents employment but does not end an in-progress court visit — so a recruited guest becomes the adventurer's courtier while still physically visiting the original court, and never travels with them.

Fix: follow `add_courtier` with `return_to_court = yes` on the chosen character, inside `hidden_effect`. `effects.log` documents `return_to_court` as "returns the scope character to the employers court", so it must run in the recruit's scope — and since `add_courtier` runs first, the employer is already the adventurer.

This mirrors the vanilla idiom used everywhere a guest/traveller is moved into a court:
- `00_debug_interactions.txt:323` — `add_courtier` then `scope:recipient = { hidden_effect = { return_to_court = yes } }` (identical shape)
- `00_debug_interactions.txt:3561` — spawn-courtier
- `00_courtier_and_guest_interactions.txt:718` — `invite_to_court` `on_accept`

The effect is shared by all three option branches (`courtier_1/2/3`), so the single edit covers every path. `hidden_effect` keeps the option tooltip unchanged.

I checked whether the recruit also needs attaching to the adventurer's travel plan and concluded it does not: the two sibling LAAMP recruitments in this same file (`scope:recidivist`, `scope:hopeful`) use plain `add_courtier` with only a location fix, and `add_to_court_and_entourage_effect`'s `add_companion` branch is activity-specific. `return_to_court` places the recruit at their new employer's court, which is what `scope:hopeful`'s explicit `set_location` achieves manually.

Not chosen: filtering guests out of the candidate pool. The event's fiction is explicitly about "courtiers or guests", so that would remove content rather than fix the transfer.

Please verify in-game that a recruited **guest** now travels with the adventurer, and check `error.log` in observer mode (`return_to_court` logs "cannot return, not a courtier" if a character has no liege — ordering here should prevent that).


events/scheme_events/laamp_base_contract_scheme_events.txt
```diff
@@ -2332,6 +2332,9 @@ scripted_trigger laamp_base_contract_schemes_0321_valid_impressionable_courtier_
 
 scripted_effect laamp_base_contract_schemes_0321_transfer_chosen_courtier_effect = {
 	add_courtier = $COURTIER_CHOSEN$
+	$COURTIER_CHOSEN$ = { #Unop: The chosen character can be a guest, and add_courtier alone does not end an ongoing court visit, so they would stay at the original court instead of travelling with us
```
pharaox: **reviewer:** `Low` (cosmetic, non-blocking) — this marker runs ~150 chars, the longest in the file. The reason recorded is exactly the right one; just a bit long for the project's marker style:

```suggestion
	$COURTIER_CHOSEN$ = { #Unop The chosen character can be a guest, and add_courtier alone does not end an ongoing court visit
```

Entirely optional.

For the record, the `=` here (rather than `?=` as used by `$REJECTED_1$`/`$REJECTED_2$` below) is correct: options B and C are gated on `exists = scope:courtier_2/3` and option A is unconditional, so `$COURTIER_CHOSEN$` always exists — matching the `add_courtier = $COURTIER_CHOSEN$` line directly above.

pharaox (commented):
**reviewer:** Looks good — minimal, mirrors an exact vanilla precedent, and the scope detail is right.

What I checked beyond the diff:

- **`return_to_court` semantics verified directly in `effects.log:9116`** rather than taken on trust: *"Returns the scope character to the employers court. Supported Scopes: character"*. That confirms it must run in the **character's** scope, so the `$COURTIER_CHOSEN$ = { ... }` wrapper is correct. Good catch in the fix notes that the `07_dlc_ep3_scripted_effects.txt:1762` precedent runs it in the *liege's* scope and was the wrong model to copy.
- **Precedents re-checked.** `00_debug_interactions.txt:322-327` is structurally identical to this diff — `add_courtier` first, then `scope:recipient = { hidden_effect = { return_to_court = yes } }`. Also confirmed `00_debug_interactions.txt:3561` and `00_courtier_and_guest_interactions.txt:716-721`. The ordering assumption (recruit already has an employer, so they return to the *adventurer's* court) is well corroborated.
- **Pattern class re-enumerated independently.** Of the four `add_courtier` sites in this file, only line 2334 draws its character from a `courtier_or_guest` iterator — `scope:recidivist` (:10026), `scope:hopeful` (:15809) and `scope:beautiful_peasant` (:29378) do not. The other two `courtier_or_guest` iterators (0551, 0561) contain **no** `add_courtier` / `return_to_court` / `set_location` anywhere in their range, so they genuinely never transfer anyone and are correctly left alone.
- **Premise confirmed:** zero occurrences of `is_pool_guest` / `is_foreign_court_guest` / `is_foreign_court_or_pool_guest` in the file, so guests really do reach the transfer unfiltered.
- **Coverage:** all three option branches (:2513/:2532/:2551) route through the shared effect. Option A is unconditional and B/C are gated on `exists`, so `$COURTIER_CHOSEN$` is always present — `=` rather than `?=` is correct here and consistent with the `add_courtier` line above, while the `$REJECTED_N$` blocks rightly keep `?=`.
- **No regression for the ordinary-courtier case:** both `invite_to_court` and the debug interaction apply `return_to_court` unconditionally to every recruit, not just guests; for a non-travelling courtier it is a no-op relocation to what is already their employer's court.
- **Tiger re-run independently:** `fatal: 0, error: 0, warning: 0, untidy: 0, tips: 0`. No suppressions added. Single commit, one file, +3 lines.

Agreed with choosing this over filtering guests out of the candidate pool — the event's loc is explicitly about "courtiers **or** guests", so filtering would remove content to hide a transfer defect.

Only a `Low` note on the marker's length (inline). Worth an in-game confirmation that a recruited **guest** now travels with the adventurer.

PR #660, opened 2026-09-21
Fix Crime Pays doubling transport contract rewards

### LINKED ISSUE #652: "Crime Pays" doubles also the transport contract rewards, not just criminal ones
"Crime Pays" perk (`malice_implicit_perk`) should only double criminal contract rewards, as its description is "[adventurer_i] [criminal_task_contract|E] [gold|E] Rewards: [gold_i]#P +100%#!". But it also doubles the transport contract rewards, as it's used in `task_contract_success_gold_gain_travel_value`.

pharaox:
Fixes #652

**fixer:** `task_contract_success_gold_gain_travel_value` contains a `has_perk = malice_implicit_perk → multiply = 2` block that is identical to the one in `task_contract_success_gold_gain_crime_value` — a copy-paste from when the travel family was derived from the criminal one. So "Crime Pays" doubles transport contract rewards, which its tooltip never claims.

Fix: comment out that block in `_travel_value` only, leaving its legitimate `at_any_cost_perk` check intact.

Evidence the criminal scoping is intentional, not incidental:
- **The generic family is the reference shape.** `task_contract_success_gold_gain_full_value` has the `at_any_cost_perk` block and *no* `malice_implicit_perk` block. So the block's presence in `_travel_value` is the anomaly, not `_crime_value`'s legitimacy.
- **The two perks' loc differ exactly where the code should.** `adventurer_at_any_cost_perk_effect` says `[task_contract|E]` with an explicit "(Not for [mercenary_task_contracts|E])" carve-out; `adventurer_malice_implicit_perk_effect` says `[criminal_task_contract|E]` with no carve-out. Vanilla writing a caveat for one and not the other shows the scoping is maintained deliberately.
- **Consumers are cleanly separated.** `_travel_value` is used only by `laamp_transport_contracts.txt`, `_crime_value` only by `laamp_base_contracts.txt`; `transport_task_contract` and `criminal_task_contract` are distinct game concepts.
- **Magnitude corroborates it.** Criminal contracts have the highest base (100 vs travel 75 vs generic 50), and `_crime_value`'s own comment says "we still want criminal contracts to almost always pay more" — doubling travel too erodes that premium.

The derived variants (`_critical_travel_value`, `_half_travel_value`) all begin with `value = task_contract_success_gold_gain_travel_value`, so they inherit the fix; no separate edits needed. The identical-looking block in `_crime_value` is correct and untouched — since the two blocks are identical I edited by line number to avoid hitting the wrong one.

Not chosen: rewording the loc to match the code. That reads as a balance change rather than a fix, contradicts the stated criminal-premium intent, and would leave every non-English string wrong (Unop ships English only).

Please verify in-game that a transport contract's reward breakdown no longer shows the Crime Pays ×2, while a criminal contract still does.


common/script_values/07_ep3_values.txt
```diff
@@ -2940,12 +2940,12 @@ task_contract_success_gold_gain_travel_value = {
...
-				scope:task_contract.task_contract_taker = { has_perk = malice_implicit_perk }
-			}
-			multiply = 2
-		}
+		#if = { #Unop: Crime Pays only doubles criminal contract rewards, not transport ones
+			#limit = {
+				#scope:task_contract.task_contract_taker = { has_perk = malice_implicit_perk }
+			#}
+			#multiply = 2
+		#}
```
pharaox:
```suggestion
		#Unop: Crime Pays only doubles criminal contract rewards, not transport ones
		#if = {
		#	limit = {
		#		scope:task_contract.task_contract_taker = { has_perk = malice_implicit_perk }
		#	}
		#	multiply = 2
		#}
```

common/script_values/07_ep3_values.txt
```diff
@@ -2940,12 +2940,13 @@ task_contract_success_gold_gain_travel_value = {
 			}
 			multiply = 1.25
 		}
-		if = {
-			limit = {
-				scope:task_contract.task_contract_taker = { has_perk = malice_implicit_perk }
-			}
-			multiply = 2
-		}
+		#Unop: Crime Pays only doubles criminal contract rewards, not transport ones
```
pharaox: **reviewer:** `Low` (non-blocking, history only) — the commit that produced this restyle is titled `Update common/script_values/07_ep3_values.txt`, the GitHub web-editor default, rather than a short imperative summary as `CONTRIBUTING.md` asks for.

Worth squashing it into `Fix Crime Pays doubling transport contract rewards` before merge so the history reads cleanly. No change needed to the code itself — the content and placement here are correct, and the block commented out is verifiably the travel one, not the criminal one.

pharaox (commented):
**reviewer:** Looks good — correct block removed, and the consumer mapping proves criminal rewards are untouched.

The main risk here was hitting the **wrong** block, since the travel and crime perk checks are byte-identical and the edit was made by line number. Confirmed correct by family boundaries:

- the commented-out `malice_implicit_perk` sits at :2946, between `_travel_value` (:2930) and the next family (:2980);
- the **live** one is at :3029, inside `_crime_value` (:3014).

I read `_crime_value` in full — base 100 with both `at_any_cost_perk` ×1.25 and `malice_implicit_perk` ×2 fully intact.

Also checked beyond the diff:

- **`_travel_value` now matches the reference shape.** It is structurally identical to `_full_value` (:2845), which carries `at_any_cost_perk` and no `malice_implicit_perk` — confirming the block's presence in the travel family was the anomaly, not `_crime_value`'s legitimacy.
- **The key assumption is stronger than stated.** I grepped the travel family across *all* of vanilla `common/` + `events/` **and** all of Unop: the sole consumer is `laamp_transport_contracts.txt` (9 call sites), plus the defining file. So this provably cannot affect criminal payouts.
- **Design intent verified from the loc itself** (`lifestyle_adventurers_l_english.yml:6` / `:32`): `adventurer_at_any_cost_perk_effect` uses generic `[task_contract|E]` *with* an explicit `(Not for [mercenary_task_contracts|E])` carve-out, while `adventurer_malice_implicit_perk_effect` names `[criminal_task_contract|E]` with none. Vanilla writing a caveat for one and not the other shows the scoping is deliberate — so aligning code to loc is the right direction, and no loc change is needed.
- **Derived variants inherit the fix:** `_critical_travel_value` (:2981) and `_half_travel_value` (:3008) both open with `value = task_contract_success_gold_gain_travel_value`.
- **Tiger re-run independently:** `fatal: 0, error: 0, warning: 0, untidy: 0, tips: 0`. `git diff --check` is clean too — no whitespace or line-ending damage, which was a stated concern given the binary-mode edit.

On the marker style: the second commit restyled it from `#if = { #Unop: ...` (the form used in `02_vassal_values.txt:206` and `00_court_position_values.txt:6322`) to a marker line above the block with preserved inner indentation. Since both commits are yours, I read that as a deliberate readability call and am not raising it as a finding.

Only a `Low` on the second commit message (inline). Worth an in-game check that a transport contract's reward breakdown no longer shows the ×2, while a criminal one still does.

PR #661, opened 2026-09-21
Fix refound_pandidakterion_decision AI gate checking gold instead of treasury

### LINKED ISSUE #640: refound_pandidakterion_decision checks gold only, ignoring the treasury it can also use
**Reported on Steam by `White Lich` on `2026-08-28 19:37`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> refound_pandidakterion_decision also uses only uses gold as the check even though it can use treasury.

pharaox:
Fixes #640

**fixer:** `refound_pandidakterion_decision` gates the AI on `gold > 1000`, but its `is_valid` requires `government_has_flag = government_is_administrative`, and `administrative_government` declares both `treasury = yes` and `replace_gold_cost_by_treasury = yes`. Per `_governments.info`, that flag means "this government pays for state expenses with treasury instead of gold" — so every actor who can legally take this decision pays from treasury, and the AI is gated on a resource it doesn't accumulate for state expenses.

Fix: comment out the `gold > 1000` line.

The decision's own `cost` block is the decisive evidence that this is an oversight rather than intent — it already branches on `has_treasury`, declaring gold **and** treasury with mutually exclusive guards. The author clearly knew payment comes from treasury; `ai_potential` just wasn't updated to match.

Dropping the check loses nothing, because `cost` is engine-enforced against the correct resource (`_decisions.info` describes `minimum_cost` as "like 'cost', except this doesn't get applied" and blocks taking the decision when unaffordable — confirming plain `cost` both applies and blocks). The gold line therefore added no safety; it only subtracted availability.

I scanned all 61 vanilla decisions with a `treasury` cost block to establish the convention and check for siblings:

| `ai_potential` | Count |
|---|---|
| No resource check (lets `cost` do the work) | 44 |
| No `ai_potential` at all | 10 |
| `short_term_gold` check | 5 |
| `gold` check | 2 |

Crucially, of the 7 that do gate on a resource, **`refound_pandidakterion_decision` is the only one restricted to administrative government** — the other six (`fp3_sponsor_research_project`, `council_of_toledo`, `form_bosporan_kingdom`, `improve_city_province`, `raise_stele`, `secure_mediterranean`) have no government gate, so their gold check still applies to genuine gold-users and is not a bug. No sibling needs the same fix.

Not chosen: making the check treasury-aware with a `has_treasury` branch mirroring `cost`. Since `is_valid` already restricts to administrative, the gold arm would be dead code — so simply removing the check is equivalent in practice and matches what 54 of 61 comparable decisions do.

Please verify in observer mode with an AI Byzantine administrative emperor holding Constantinople and ≥1000 treasury that the decision now gets taken. Note `ai_check_interval_by_tier` is 120 months at empire/hegemony tier, so this needs a long run or a targeted save.


common/decisions/dlc_decisions/ep3_decisions.txt
```diff
@@ -658,7 +658,7 @@ refound_pandidakterion_decision = {
 
 	ai_potential = {
 		has_title = title:e_byzantium
-		gold > 1000
+		#gold > 1000 #Unop: This decision requires an administrative government, which pays from treasury instead of gold, so the AI could never meet this; cost already enforces affordability against the right resource
```
pharaox:
```suggestion
		#gold > 1000 #Unop: Admin govs pay from treasury instead of gold; cost already enforces this
```

common/decisions/dlc_decisions/ep3_decisions.txt
```diff
@@ -658,7 +658,7 @@ refound_pandidakterion_decision = {
 
 	ai_potential = {
 		has_title = title:e_byzantium
-		gold > 1000
+		#gold > 1000 #Unop: Admin govs pay from treasury instead of gold; cost already enforces this
```
pharaox: **reviewer:** The marker placement and wording are right — on the commented line, per the comment-out convention, and the shortened text reads well (the original ~190-char version was unwieldy).

`Low` (non-blocking, history only): the commit that made this shortening is titled `Update common/decisions/dlc_decisions/ep3_decisions.txt`, the GitHub web-editor default, rather than a short imperative summary as `CONTRIBUTING.md` asks for. Worth squashing it into `Fix refound_pandidakterion_decision AI gate checking gold instead of treasury` before merge.

No code change needed.

pharaox (commented):
**reviewer:** Looks good — the gate was unsatisfiable by construction, and removing it is the conventional remedy.

Verified the decisive claim from the authoritative source rather than on trust:

- `00_government_types.txt:425-426` — `administrative_government` sets both `treasury = yes` and `replace_gold_cost_by_treasury = yes`.
- `_governments.info:228-230` documents that flag as *"This government pays for state expenses with treasury instead of gold."*
- `is_valid` (:609) hard-requires `government_is_administrative`.

So **every** legal taker pays from treasury, and `gold > 1000` could never be the right resource. The `cost` block (:625-644) already branches on `has_treasury` with mutually exclusive guards — the author knew this; `ai_potential` just wasn't updated. Local inconsistency, not intent.

Also checked:

- **The "does removal lose safety?" assumption.** `_decisions.info:134` documents `minimum_cost` as *"Like 'cost', except this **doesn't get applied**. It'll block taking the decision if not affordable"* — establishing by contrast that plain `cost` both applies **and** blocks. Affordability stays engine-enforced against the correct resource.
- **A pitfall that could have made this a no-op.** CK3 has a known behaviour where the AI won't take a decision with an *influence* cost unless `ai_goal = yes`. If something analogous applied to treasury, dropping the gold gate wouldn't actually unblock the AI. It doesn't: **zero** vanilla decision files contain both a `treasury` cost and `ai_goal`, and `_decisions.info:107` describes `ai_goal` as a budgeting/interval mechanism for major costs, not a hard gate.
- **Convention re-verified independently:** `30_court_decisions.txt` (treasury-using) has zero gold/treasury checks across all its `ai_potential` blocks.
- **Sibling check — no missed member.** The only other `gold >= 1000` AI gate in vanilla decisions is `start_city_construction_decision` (`06_ep3_laamp_decisions.txt:2863`), gated to `government_is_landless_adventurer` — a real gold-user, so its check is correct and not a sibling defect.
- **Tiger re-run independently:** `fatal: 0, error: 0, warning: 0, untidy: 0, tips: 0`. `git diff --check` clean.

Agreed with dropping the line rather than writing a treasury-aware check — with `is_valid` restricting to administrative government, a `has_treasury = no` arm would be dead code.

Only a `Low` on the second commit message (inline). In-game confirmation is worthwhile but slow: `ai_check_interval_by_tier` is 120 months at empire/hegemony, so it needs a targeted save rather than a casual observer run.

PR #662, opened 2026-09-21
Fix demand_admin_interaction AI wealth check being inverted and gold-only

### LINKED ISSUE #639: demand_admin_interaction gold requirements are flipped, so AI only demands when it lacks gold; also checks gold instead of treasury
**Reported on Steam by `White Lich` on `2026-08-28 19:37`** (https://steamcommunity.com/sharedfiles/filedetails/?id=2871648329).

> AI will almost never use the demand_admin_interaction because the gold requirements are flipped so it will only demand when they lack gold. Also uses gold as the check even though it can use treasury.

pharaox:
Fixes #639

**fixer:** Two independent defects in `demand_admin_interaction`'s `ai_will_do`, both fixed by the same three lines.

1. **The comparison was inverted.** `factor = 0` *zeroes* the AI's desire, so `gold >= 300` collapsed the score to 0 exactly when the AI could afford the interaction, leaving it at 100 only when poor. Since the interaction costs the actor money, wealth should enable it, not veto it.
2. **It read the wrong resource.** `ai_potential` restricts the interaction to `government_allows = administrative`, and four eligible governments (`administrative`, `celestial`, `steppe_admin`, `meritocratic`) declare `replace_gold_cost_by_treasury = yes`. `change_to_administrative_interaction_effect` branches on `has_treasury` and calls `pay_short_term_treasury` accordingly — so most eligible actors pay from treasury, while the gate only ever looked at gold. (`japan_administrative_government` is the one eligible government without treasury, so both branches are live.)

Fix: replace each `gold >= X` with `treasury_or_gold < X`. `triggers.log` documents `treasury_or_gold` as "Does the character replace gold cost by treasury? Compare value against treasury, otherwise gold" — a purpose-built trigger that expresses the resource split in one line and exactly mirrors what the paying effect does. It's an established vanilla idiom, used in this comparison form at `00_laamp_triggers.txt:485` (`treasury_or_gold >= 50`), alongside `pay_treasury_or_gold` and `cost = { treasury_or_gold = ... }`.

I kept the existing tier structure and the 300/600/2000 thresholds. The real costs are lower (`50 ×` tier multiplier = 150/300/500 before add-ons), so these now act as conservative affordability floors — correcting the two defects without turning the fix into a rebalance.

**I deliberately did not take the structural route** of giving the interaction `cost = { gold / treasury }` blocks the way `feudalize_holding_interaction` does. `change_to_administrative_interaction_effect` has **eight other callers** in `ep3_interactions_events.txt`, and the payment happens there — `on_accept` only wraps it in `show_as_tooltip`. Moving payment into a `cost` block would double-charge those paths or require touching all nine call sites, which is far beyond the scope of this bug.

Please verify in observer mode with administrative AI realms that AI lieges now demand administrative government from vassals at a plausible rate, and confirm `error.log` shows no new `pay_short_term_treasury` failures.


common/character_interactions/06_ep3_interactions.txt
```diff
@@ -2735,17 +2735,17 @@ demand_admin_interaction = {
 		modifier = {
 			factor = 0
 			scope:recipient = { highest_held_title_tier >= tier_county }
-			gold >= 300
+			treasury_or_gold < 300 #Unop: This was inverted, zeroing the desire when we could afford it, and only checked gold although administrative governments pay from treasury
```
pharaox:
```suggestion
			treasury_or_gold < 300 #Unop: This was inverted, and only checked gold
```

common/character_interactions/06_ep3_interactions.txt
```diff
@@ -2735,17 +2735,17 @@ demand_admin_interaction = {
...
 		}
 		modifier = {
 			factor = 0
 			scope:recipient = { highest_held_title_tier >= tier_duchy }
-			gold >= 600
+			treasury_or_gold < 600 #Unop: This was inverted, zeroing the desire when we could afford it, and only checked gold although administrative governments pay from treasury
```
pharaox:
```suggestion
			treasury_or_gold < 600 #Unop: This was inverted, and only checked gold
```

common/character_interactions/06_ep3_interactions.txt
```diff
@@ -2735,17 +2735,17 @@ demand_admin_interaction = {
...
 		}
 		modifier = {
 			factor = 0
 			scope:recipient = { highest_held_title_tier >= tier_kingdom }
-			gold >= 2000
+			treasury_or_gold < 2000 #Unop: This was inverted, zeroing the desire when we could afford it, and only checked gold although administrative governments pay from treasury
```
pharaox:
```suggestion
			treasury_or_gold < 2000 #Unop: This was inverted, and only checked gold
```

pharaox (commented):
**reviewer:** Looks good — no findings. Both bugs are fixed by the same three lines, and `treasury_or_gold` is the right primitive.

`factor = 0` now applies when the actor is too **poor**, so wealth enables the interaction instead of vetoing it (bug 1), and the resource read is treasury-aware (bug 2). Neither bug could have been fixed alone.

Verified beyond the diff:

- **`treasury_or_gold` confirmed in `triggers.log:7203`** rather than assumed: *"Does the character replace gold cost by treasury? Compare value against treasury, otherwise gold."* — traits include `<`, scope `character`.
- **Probed a subtle divergence and cleared it.** The paying effect branches on `has_treasury`, but `treasury_or_gold` keys on `replace_gold_cost_by_treasury` — *different* flags, so a government carrying one but not the other would make the AI gate disagree with the actual payment. Checked every government file: all four treasury governments declare both (administrative 425/426, celestial 744/746, steppe_admin 1037/1038, meritocratic 1162/1163). The two selectors pick the same resource everywhere, so there is no gap.
- **Eligibility enumerated:** `government_allows = administrative` matches exactly five governments — those four plus `japan_administrative_government` (Ritsuryo), which has no treasury. So both arms of `treasury_or_gold` stay live and the gold arm correctly serves Ritsuryo.
- **Cost magnitudes confirmed** against `change_to_administrative_interaction_effect` (`00_administrative_effects.txt:295`): `50 ×` tier multiplier → 500 / 300 / 150. The retained 300/600/2000 thresholds sit above these, so once inverted they act as conservative affordability floors. Agreed with keeping the vanilla numbers — re-deriving them would turn a correctness fix into a rebalance.
- **Remedy 2 correctly ruled out, and I checked the evidence.** The effect has eight call sites — this `on_accept` plus seven in `ep3_interactions_events.txt` (32, 95, 135, 204, 246, 284, 326) — and `on_accept` (:2692-2695) only wraps it in `show_as_tooltip`, so real payment happens in the event handlers. Moving payment into a `cost` block would double-charge those paths.
- **Idiom precedent:** `treasury_or_gold` appears in this exact comparison form at `00_laamp_triggers.txt:485` and across ten+ vanilla files including `00_tribal_interactions.txt`. Preferring it over a hand-rolled `has_treasury` branch is simpler and semantically exact.
- **Tiger re-run independently:** `fatal: 0, error: 0, warning: 0, untidy: 0, tips: 0` — which also validates `treasury_or_gold` as a real trigger name. `git diff --check` clean.

The markers sit on the changed lines, which is right for in-place value changes, and your batched suggestions already tightened them.

Worth an observer-mode run with administrative AI realms to confirm lieges now demand administrative government at a plausible rate, and that `error.log` shows no new `pay_short_term_treasury` failures.

